PR #9, opened 2026-02-26
Fix clippy warnings and WhatsApp gateway Node 25 compat

devatsecure:
## Summary
- Fix 2 collapsible else-if clippy warnings in CLI `doctor` command
- Add `"type": "commonjs"` to `packages/whatsapp-gateway/package.json` — Node.js 25 defaults to ESM and the gateway uses `require()`
- Replace `import.meta.url` with `__dirname` in `packages/whatsapp-gateway/index.js` — `import.meta` is invalid in CommonJS context

## Test plan
- [x] `cargo clippy --workspace --all-targets -- -D warnings` passes with zero warnings
- [x] `cargo test --workspace` — 1,769 tests pass
- [x] WhatsApp gateway starts successfully on Node.js 25 and returns QR code
- [x] Dashboard serves correctly at `http://127.0.0.1:50051/`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3852,6 +3852,7 @@ dependencies = [
  "dirs 6.0.0",
  "futures",
  "hex",
+ "libc",
  "openfang-channels",
  "openfang-extensions",
  "openfang-hands",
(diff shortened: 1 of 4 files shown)


jaberjaber23:
Good catch on the Node 25 ESM default change. We'll patch this ourselves. Thanks.

PR #21, opened 2026-02-26
feat: add theme switcher with system preference support

felipeorlando:
## What

Replaces the binary light/dark toggle with a 3-option theme switcher (Light / System / Dark) that respects OS-level preference and persists the user's choice across reloads.

## Changes

- **`theme-manager.js`** — new utility handling persistence, system preference detection via `prefers-color-scheme`, and DOM updates. Wires a `matchMedia` listener so the UI reacts automatically when the OS theme changes (only when set to System).
- **`app.js`** — integrated `ThemeManager`: init on load, new `setTheme(t)` method accepting `'light' | 'dark' | 'system'`, legacy `toggleTheme()` kept for backward compat.
- **`index_body.html`** — replaced single emoji button with a pill-style 3-button group (☀ / ○ / ☾), active state tracked via Alpine.js binding.
- **`theme.css`** — added `.theme-switcher` and `.theme-opt` styles (pill group, active highlight with accent color).
- **`webchat.rs`** — registered `theme-manager.js` in the `include_str!` bundle before `app.js`.

## Behavior

| Scenario | Result |
|---|---|
| First visit | Defaults to System (follows OS) |
| User picks Light/Dark | Persisted to `localStorage`, stays on reload |
| User picks System | Follows OS; updates live if OS theme changes |
| OS switches while on System | UI updates automatically, no reload needed |

## Zero dependencies

Pure vanilla JS + CSS. No new build steps, no external libs.

### DIFF
--- a/crates/openfang-api/src/webchat.rs
+++ b/crates/openfang-api/src/webchat.rs
@@ -92,6 +92,8 @@ const WEBCHAT_HTML: &str = concat!(
     "<script>\n",
     include_str!("../static/js/api.js"),
     "\n",
+    include_str!("../static/js/theme-manager.js"),
+    "\n",
     include_str!("../static/js/app.js"),
     "\n",
     include_str!("../static/js/pages/overview.js"),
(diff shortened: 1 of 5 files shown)


jaberjaber23:
Love the idea, we'll add a proper theme switcher natively. Thanks for the inspiration.

PR #34, opened 2026-02-27
Propagate base_url from deployment manifest with default fallback

neoz:
The base_url in config.toml not working. This patch fix that problem
[default_model]
provider = "anthropic"                    # "anthropic", "gemini", "openai", "groq", "ollama", etc.
model = "claude-sonnet-4-20250514"        # Model identifier
api_key_env = "ANTHROPIC_API_KEY"         # Environment variable holding API key
base_url = "https://abcd.com"  # Optional: override API endpoint <---


jaberjaber23:
Both base_url propagation fixes already exist on main in more robust forms with guard checks. Thanks for identifying the issue.

PR #45, opened 2026-02-27
fix(cli): include daemon api_key auth for protected endpoints

Jungley8:


Jungley8:
Close it first, and then submit the PR after the test passes.

Jungley8:
#51

PR #62, opened 2026-02-28
ChatGPT Plus/Pro Subscription Support Using Codex authentication

cxnmai:
ChatGPT subscriptions offer token usage that comes from the same pool as codex coding agent. I've added support for that in openfang.
In init setup, "ChatGPT Sub" is now an option:
image
When clicked, user clicks enter to open the link in browser:
image
Proceed with login normally, until: 
image
The rest proceeds as normal. Models are accessed using codes like:
image
Access tokens are also configurable here as normal:
image


jaberjaber23:

PR #75, opened 2026-02-28
fix(installer): robust shell detection and fish-safe PATH update

Dwsy:
## Summary
- add robust shell detection for installer using priority: `OPENFANG_SHELL` -> `$SHELL` -> parent process -> grandparent process
- fix fish PATH injection to use `fish_add_path` instead of POSIX `export PATH=...` syntax
- keep bash/zsh/sh behavior unchanged (`export PATH=...`)
- fallback profile handling for unknown shells (`~/.profile`)

## Why
When users run `curl -fsSL https://openfang.sh/install | sh` from fish, shell detection can be ambiguous and PATH lines must be fish-native. Using POSIX export syntax in fish can break PATH and hide core commands.

## Validation
- `bash -n scripts/install.sh`
- fish context check with parent-shell fallback (`fish -c`, `SHELL=/bin/sh`)
  - detected shell = `fish`
  - profile = `~/.config/fish/config.fish`
- explicit override checks (`OPENFANG_SHELL=fish|zsh`)
- end-to-end local install test with isolated HOME:
  - fish writes `fish_add_path "<install_dir>"` to fish config
  - bash writes `export PATH="<install_dir>:$PATH"` to bash profile

## Scope
No platform download/extract logic changed. Linux/macOS target resolution remains untouched.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -38,6 +38,70 @@ detect_platform() {
     esac
 }
 
+normalize_shell_name() {
+    case "$1" in
+        *fish*) echo "fish" ;;
+        *zsh*) echo "zsh" ;;
+        *bash*) echo "bash" ;;
+        *sh) echo "sh" ;;
+        *) echo "" ;;
+    esac
+}
+
+get_process_name() {
+    local pid="$1"
+    ps -p "$pid" -o comm= 2>/dev/null | awk '{print $1}'
+}
+
+detect_user_shell() {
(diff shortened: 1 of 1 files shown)


jaberjaber23:
Fixed in d237ecf. Improved shell detection with /etc/passwd fallback for minimal containers, added config file existence checks, and switched fish PATH to use fish_add_path instead of set -gx.

PR #102, opened 2026-02-28
feat: support ?token= query param for remote browser access

Gillinghammer:
## Summary
- Reads `?token=` from the URL on dashboard page load and sets it as the API auth token
- Saves the token to `localStorage` so subsequent page loads don't need it in the URL
- Strips the token from the URL bar via `history.replaceState` to avoid leaking in bookmarks/history

## Problem
When accessing the dashboard remotely (e.g., `http://192.168.1.x:4200`), the API endpoints require a Bearer token but the browser has no way to inject `Authorization` headers on the initial page load. The backend middleware already supports `?token=` on all HTTP endpoints (middleware.rs L133-147), and the frontend already uses `?token=` for WebSocket connections (api.js L215), but there was no way to bootstrap the token into the frontend on first load.

## Solution
On `alpine:init`, check `window.location.search` for a `token` parameter before falling back to `localStorage`. This allows remote access via:
```
http://192.168.1.x:4200/?token=YOUR_API_KEY
```

The token is immediately persisted to `localStorage` and stripped from the visible URL.

## Test plan
- [ ] Access dashboard locally at `http://localhost:4200` — should work as before (no token needed)
- [ ] Access dashboard remotely without token — should show auth prompt or fail as before
- [ ] Access dashboard remotely with `?token=VALID_KEY` — should load successfully
- [ ] After loading with `?token=`, verify URL bar no longer contains the token
- [ ] Refresh page after initial `?token=` load — should still work (token in localStorage)
- [ ] Access with `?token=WRONG_KEY` — should fail auth

### DIFF
--- a/crates/openfang-api/src/middleware.rs
+++ b/crates/openfang-api/src/middleware.rs
@@ -82,6 +82,11 @@ pub async fn auth(
     // Public endpoints that don't require auth (dashboard needs these)
     let path = request.uri().path();
     if path == "/"
+        || path == "/favicon.ico"
+        || path == "/logo.png"
+        || path.starts_with("/css/")
+        || path.starts_with("/js/")
+        || path.starts_with("/vendor/")
         || path == "/api/health"
         || path == "/api/health/detail"
         || path == "/api/status"


jaberjaber23:

PR #105, opened 2026-03-01
Fix UTF-8 panic in string truncation (kernel, CLI, channels)

### LINKED ISSUE #104: Panic: UTF-8 byte boundary error in string truncation (session.rs / kernel.rs)
## Bug Description

OpenFang panics when truncating agent response strings at a byte index that falls inside a multi-byte UTF-8 character (e.g., em dash `—` which is 3 bytes: `0xE2 0x80 0x94`).

## Panic Locations

Two confirmed sites:

1. **`crates/openfang-memory/src/session.rs:433`**
2. **`crates/openfang-kernel/src/kernel.rs:418`**

Both are performing string slicing at a fixed byte offset without checking for a char boundary.

## Panic Message

```
thread 'tokio-runtime-worker' panicked at crates/openfang-memory/src/session.rs:433:51:
byte index 200 is not a char boundary; it is inside '—' (bytes 199..202) of `Here's your VPS summary — 109.123.234.207: ...`
```

```
thread 'tokio-runtime-worker' panicked at crates/openfang-kernel/src/kernel.rs:418:17:
byte index 500 is not a char boundary; it is inside '—' (bytes 499..502) of `**100 trades retrieved** ...`
```

## Impact

- The panic kills the tokio worker thread handling the agent loop
- The HTTP response for the `/api/agents/{id}/message` endpoint returns an empty reply or 500
- The agent becomes unresponsive until the daemon is restarted
- Subsequent messages to the same agent also fail (stuck state)

## Root Cause

Rust string slicing (`&s[..N]`) panics if `N` is not on a UTF-8 char boundary. Agent responses frequently contain multi-byte characters like em dashes (`—`), smart quotes, and emoji, making this a common occurrence.

## Suggested Fix

Replace direct byte slicing with a char-boundary-safe truncation. For example:

```rust
// Before (panics on multi-byte chars):
let truncated = &response[..200];

// After (safe):
fn truncate_str(s: &str, max_bytes: usize) -> &str {
    if s.len() <= max_bytes {
        return s;
    }
    // Find the last char boundary at or before max_bytes
    let mut end = max_bytes;
    while end > 0 && !s.is_char_boundary(end) {
        end -= 1;
    }
    &s[..end]
}
```

Or use the `floor_char_boundary` method (stable since Rust 1.82):

```rust
let truncated = &response[..response.floor_char_boundary(200)];
```

## Reproduction

1. Send an agent a task that produces a response containing multi-byte UTF-8 characters (em dashes, emoji, etc.) near byte positions 200 or 500
2. The daemon panics when compacting the session or processing the kernel response

## Environment

- OpenFang version: 0.1.9
- OS: macOS (aarch64, Darwin 25.3.0)
- Provider: OpenRouter
- Model: anthropic/claude-sonnet-4.6

AL-ZiLLA:
## Summary

- Replace 3 remaining raw byte-index string slices (`&s[..N]`) with the existing UTF-8-safe `openfang_types::truncate_str()` 
- Add regression test for em dash (`—`), the exact character that triggered production panics

## Problem

`&s[..200]` panics when byte 200 falls inside a multi-byte UTF-8 character. Agent responses frequently contain em dashes (3 bytes: `0xE2 0x80 0x94`), smart quotes, and emoji, making this a common crash in production.

**Panic sites fixed:**
| File | Line | Context |
|------|------|---------|
| `kernel.rs` | 2215 | Session summary truncation (`&t[..200]`) |
| `init_wizard.rs` | 1916 | Model name display (`&short[..14]`) |
| `nostr.rs` | 168 | Pubkey log message (`&pubkey[..16]`) |

## Production panic logs

```
thread 'tokio-runtime-worker' panicked at crates/openfang-kernel/src/kernel.rs:418:17:
byte index 500 is not a char boundary; it is inside '—' (bytes 499..502)
```

```
thread 'tokio-runtime-worker' panicked at crates/openfang-memory/src/session.rs:433:51:
byte index 200 is not a char boundary; it is inside '—' (bytes 199..202)
```

## Fix

All 3 sites now use `openfang_types::truncate_str(s, max_bytes)` which walks back to the nearest `is_char_boundary()`. This function already existed and was used in ~30 other call sites — these 3 were simply missed.

## Test plan

- [x] `cargo build --workspace --lib` — compiles clean
- [x] `cargo test --workspace` — 271 passed, 0 failed
- [x] `cargo clippy -p openfang-kernel -p openfang-types -p openfang-channels -p openfang-memory` — 0 warnings
- [x] New test `truncate_str_em_dash` covers the exact production failure case

Fixes #104

### DIFF
--- a/crates/openfang-channels/src/nostr.rs
+++ b/crates/openfang-channels/src/nostr.rs
@@ -165,7 +165,7 @@ impl ChannelAdapter for NostrAdapter {
     ) -> Result<Pin<Box<dyn Stream<Item = ChannelMessage> + Send>>, Box<dyn std::error::Error>>
     {
         let pubkey = self.derive_pubkey();
-        info!("Nostr adapter starting (pubkey: {}...)", &pubkey[..16]);
+        info!("Nostr adapter starting (pubkey: {}...)", openfang_types::truncate_str(&pubkey, 16));
 
         if self.relays.is_empty() {
             return Err("Nostr: no relay URLs configured".into());
--- a/crates/openfang-cli/src/tui/screens/init_wizard.rs
+++ b/crates/openfang-cli/src/tui/screens/init_wizard.rs
@@ -1912,11 +1912,7 @@ fn draw_routing_pick(f: &mut Frame, area: Rect, state: &mut State, tier: usize) 
                 .split('/')
                 .next_back()
                 .unwrap_or(&state.routing_models[t]);
-            let display = if short.len() > 14 {
-                &short[..14]
-            } else {
-                short
-            };
+            let display = openfang_types::truncate_str(short, 14);
             summary_spans.push(Span::styled(
                 format!("{name}:{display}"),
                 Style::default().fg(*c),
--- a/crates/openfang-kernel/src/kernel.rs
+++ b/crates/openfang-kernel/src/kernel.rs
@@ -2212,7 +2212,7 @@ impl OpenFangKernel {
                 .take(5)
                 .enumerate()
                 .map(|(i, t)| {
-                    let truncated = if t.len() > 200 { &t[..200] } else { t };
+                    let truncated = openfang_types::truncate_str(t, 200);
                     format!("{}. {}", i + 1, truncated)
                 })
                 .collect::<Vec<_>>()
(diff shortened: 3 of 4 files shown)


jaberjaber23:
Implemented in v0.3.25. All 3 remaining raw byte-index slices replaced with truncate_str() plus em dash regression test.

PR #110, opened 2026-03-01
Fix CLI binary killed on macOS Apple Silicon

### LINKED ISSUE #109: CLI binary killed on macOS Apple Silicon (Killed: 9)
## Problem

The CLI binary downloaded via `curl -sSf https://openfang.sh | sh` is immediately killed on macOS Apple Silicon:

```
$ openfang --version
Killed: 9
```

Exit code 137 (SIGKILL). The system log shows:

```
kernel: load code signature error 2 for file "openfang"
(AppleSystemPolicy) ASP: Security policy would not allow process: ..., .openfang/bin/openfang
```

## Root cause

The CLI release job in `release.yml` builds macOS binaries with `cargo build` but does not codesign them. The resulting binary only has a **linker-signed** signature (the minimal signature the Apple linker adds automatically on arm64):

```
$ codesign -d -vvvv ~/.openfang/bin/openfang
Signature=adhoc
flags=0x20002(adhoc,linker-signed)
```

Recent macOS versions reject linker-signed-only binaries via Apple System Policy. This is distinct from the Desktop/Tauri builds, which are properly Developer ID signed and notarized.

## Impact

- **Affects**: all macOS Apple Silicon users installing via the CLI installer
- **Does not affect**: Desktop app (Tauri, properly signed), Docker, Linux, Windows

## Workaround

Users can re-sign the binary locally:

```bash
codesign --force --sign - ~/.openfang/bin/openfang
```

## Suggested fix

1. Add `codesign --force --sign -` to the CLI job in `release.yml` after building, before packaging (one-liner, no certificate needed)
2. Add the same to `scripts/install.sh` after extraction as a safety net

Ideally the CLI binaries should also be Developer ID signed + notarized like the Desktop builds, but ad-hoc signing is the minimum fix.

## Environment

- macOS 15 (Darwin 25.3.0)
- Apple Silicon (arm64)
- openfang 0.2.2 installed via `curl -sSf https://openfang.sh | sh`

renaudguerin:
## Summary

Fixes #109 -- CLI binary is immediately killed (SIGKILL / Killed: 9) on macOS Apple Silicon due to missing code signature.

- Add ad-hoc codesigning step to the CLI release job in release.yml, after build and before packaging
- Add ad-hoc codesigning to scripts/install.sh after extraction as a safety net

The desktop/Tauri builds are already properly Developer ID signed and notarized -- this only affects the CLI binary.

## Details

The CLI job produces binaries that are only **linker-signed** (the minimal signature Apple ld adds on arm64). Recent macOS versions reject these via Apple System Policy. Running codesign --force --sign - produces a proper ad-hoc signature that satisfies ASP without requiring a Developer ID certificate.

## Test plan

- [ ] Build a macOS arm64 CLI binary without this fix -- confirm it is linker-signed
- [ ] Build with this fix -- confirm codesign -d -vvvv shows adhoc without linker-signed
- [ ] Run the signed binary on macOS Apple Silicon -- confirm it launches without Killed: 9
- [ ] Run scripts/install.sh on macOS -- confirm codesign step runs after extraction

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -181,6 +181,9 @@ jobs:
       - name: Build CLI
         if: matrix.target != 'aarch64-unknown-linux-gnu'
         run: cargo build --release --target ${{ matrix.target }} --bin openfang
+      - name: Ad-hoc codesign CLI binary (macOS)
+        if: runner.os == 'macOS'
+        run: codesign --force --sign - target/${{ matrix.target }}/release/openfang
       - name: Package (Unix)
         if: matrix.archive == 'tar.gz'
         run: |
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -110,6 +110,11 @@ install() {
     tar xzf "$ARCHIVE" -C "$INSTALL_DIR"
     chmod +x "$INSTALL_DIR/openfang"
 
+    # macOS on Apple Silicon requires a valid code signature
+    if [ "$OS" = "darwin" ] && command -v codesign &>/dev/null; then
+        codesign --force --sign - "$INSTALL_DIR/openfang" 2>/dev/null || true
+    fi
+
     # Add to PATH
     SHELL_RC=""
     case "${SHELL:-}" in


jaberjaber23:
Implemented in v0.3.25. macOS ad-hoc codesign added to release CI and install script.

Issue #106, opened 2026-03-01
fix: PATCH /api/agents/:id does not update model or name fields


ntamero:
## Bug Description

When using the REST API to update an agent's model or name via `PATCH /api/agents/:id`, the changes are silently ignored. The response returns no error but the agent retains its original values.

## Steps to Reproduce

1. Create an agent: `openfang agent new assistant`
2. Try to update the model via API:
```bash
curl -X PATCH http://127.0.0.1:4200/api/agents/<agent-id> \
  -H "Content-Type: application/json" \
  -d '{"model": {"model": "deepseek/deepseek-chat", "provider": "openrouter"}}'
```
3. Verify: `curl http://127.0.0.1:4200/api/agents/<agent-id>`
4. The model remains unchanged.

Same behavior with `name` field — PATCH silently ignores it.

## Expected Behavior

PATCH should update the specified fields and persist the changes.

## Workaround

Currently the only way to change an agent's model is to kill it and create a new one:
```bash
openfang agent kill <id>
# Change default_model in config.toml
openfang agent new assistant
```

## Environment

- OpenFang v0.2.2
- Linux (Ubuntu 24.04, Hetzner VPS)
- Provider: OpenRouter

## Related

This is related to #101 (`feat(cli): add agent set command for live model switching`) — until `agent set` is implemented, the REST API PATCH should at least work for model changes.

manhnt9:
This is seriously needed. Currently it is not possible to spawn agent with a specific model. It all uses default model in the toml file. Editing agent toml or paste raw toml via Dashboard is not working

jaberjaber23:
Fixed in v0.2.4. Added model, provider, api_key_env, and base_url fields to PatchAgentConfigRequest. PATCH /api/agents/:id now updates model and provider.

PR #124, opened 2026-03-01
fix(api): support legacy PATCH /api/agents/{id} name/model updates

Owen718:
### Motivation
Issue #106 requires restoring backward compatibility for older clients that send partial `{ name, model }` updates to `PATCH /api/agents/{id}`.

### What changed
- Kept the API behavior focused on legacy compatibility and validated it with integration coverage.
- Updated API integration test routing for `/api/agents/{id}` to include `GET/PUT/PATCH/DELETE`, matching the runtime router behavior.
- Added `test_legacy_patch_updates_name_and_model` to verify:
  - legacy PATCH accepts `{ name, model }`
  - response returns updated values
  - values persist and are observable via subsequent `GET /api/agents/{id}`

### Contributing.md alignment
- Follows the recommendation that every new feature/behavior should include tests.
- Uses the project’s existing real-HTTP integration test pattern (temp dir + random port), consistent with existing test infrastructure.

### Validation
- `cargo test -p openfang-api --test api_integration_test --no-run`
- `cargo test -p openfang-api --test api_integration_test test_legacy_patch_updates_name_and_model`  
  (currently fails in this environment with an existing `403` on `/api/agents` spawn path, which also affects pre-existing spawn-based integration tests)

### DIFF
--- a/crates/openfang-runtime/src/agent_loop.rs
+++ b/crates/openfang-runtime/src/agent_loop.rs
@@ -57,9 +57,19 @@ const MAX_HISTORY_MESSAGES: usize = 20;
 /// but the upstream API expects just `org/model`. This also handles special routers
 /// like `openrouter/auto` → `auto`.
 pub fn strip_provider_prefix(model: &str, provider: &str) -> String {
-    let prefix = format!("{}/", provider);
-    if model.starts_with(&prefix) {
-        model[prefix.len()..].to_string()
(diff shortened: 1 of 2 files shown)


jaberjaber23:
Fixed in d237ecf. Added PATCH /api/agents/{id} endpoint supporting partial updates to name, description, model, and system_prompt. Changes persist to SQLite.

PR #217, opened 2026-03-02
Filter new-agent model list by selected provider

kocendavid:
## Summary
- load available models for the spawn wizard from '/api/models?available=true'
- filter model options by selected provider in the New Agent modal
- auto-sync selected model when provider changes
- keep manual model input fallback when provider has no available models

## Why
The Create Agent flow should only show models that are actually available for the selected provider to avoid invalid combinations and reduce user error.

## Testing
- Open Create Agent modal (wizard mode)
- In Step 2, change provider and verify model list updates to provider-specific available models
- Verify providers with no available models show manual model input fallback
- Spawn agent successfully with selected model


jaberjaber23:
Fixed in d237ecf. Provider filter dropdown added to model switcher with dynamic provider list extraction.

PR #222, opened 2026-03-02
fix(exec_policy): resolve type mismatch and approval gate ordering for per-agent exec_policy override

### LINKED ISSUE #182: exec_policy = "allow" in agent manifest [capabilities] block is silently ignored
**Version:** v0.2.7
**Component:** Agent manifest / `shell_exec` approval gate

## Description

Setting `exec_policy = "allow"` inside the `[capabilities]` block of an agent manifest TOML has no observable effect. The `shell_exec` approval gate still fires on every execution, blocking autonomous pipeline runs even when the manifest explicitly opts in.

## Steps to Reproduce

1. Create an agent manifest with:
   ```toml
   [capabilities]
   tools       = ["web_search", "web_fetch", "shell_exec", ...]
   exec_policy = "allow"
   ```
2. Spawn the agent: `openfang agent spawn manifest.toml`
3. Trigger a run that requires `shell_exec`
4. Observe: agent responds asking for human approval despite `exec_policy = "allow"` in the manifest

## Expected Behaviour

`exec_policy = "allow"` in the manifest pre-authorises shell execution for that agent — no per-run approval prompt.

## Actual Behaviour

The key is parsed without error (no warning in logs, no spawn failure) but has no effect at runtime. The agent response includes:

> "The system requires explicit human approval for shell execution, even with your autonomous authorization."

## Additional Context

- `openfang approvals list` shows an empty queue — the gate is not creating an approval item, the execution is simply blocked
- `openfang agent set <id> model <value>` exists but there is no `openfang agent set <id> exec_policy <value>` equivalent
- This makes it impossible to configure autonomous `shell_exec` for scheduled pipeline agents without either accepting interactive approval or removing `shell_exec` from the toolset entirely

capt-marbles:
Fixes #182.

## Root Cause

Two independent bugs caused `exec_policy = "allow"` in agent manifests to be silently ignored, leaving every `shell_exec` call blocked by the human approval gate even when an agent explicitly opts into full exec mode.

### Bug 1 — serde type mismatch (`agent.rs`)

`AgentManifest::exec_policy` was typed as `Option<ExecPolicy>` (a struct), but manifest authors write a bare string:

```toml
[capabilities]
exec_policy = "allow"
tools = ["shell_exec"]
```

Serde cannot deserialize a string into `ExecPolicy`, so `#[serde(default)]` silently returns `None`. The kernel then inherits the global `Allowlist` policy. This is confirmed in the kernel log from v0.3.0:

```
Agent exec_policy resolved agent=test-v30 exec_mode=Some(Allowlist)
```

`ExecSecurityMode` already has the correct aliases (`"allow"` → `Full`, `"restricted"` → `Allowlist`, `"deny"` → `Deny`), but the field type is wrong so they are never reached.

### Bug 2 — approval gate fires before exec_mode check (`tool_runner.rs`)

Even if the field parsed correctly, the approval gate in `execute_tool` fired unconditionally before any exec_policy inspection. For scheduled / headless agents there is no human in the loop, so the gate always times out and denies the tool call. The `is_full_exec` guard already existed further down in the function (for the taint check); it was just missing at the gate.

## Fix

**`crates/openfang-types/src/serde_compat.rs`** — adds `exec_policy_override` deserializer that accepts either a bare `ExecSecurityMode` string or a full `ExecPolicy` struct. A bare string is expanded into an `ExecPolicy` with that mode and all other fields at defaults. Absent / null fields return `None`, preserving backward compatibility with existing DB blobs.

**`crates/openfang-types/src/agent.rs`** — wires the new deserializer onto `AgentManifest::exec_policy` via `deserialize_with`.

**`crates/openfang-runtime/src/tool_runner.rs`** — computes `is_full_exec` before the approval gate and skips the gate when the agent is in Full exec mode.

## Tests

Six new tests in `serde_compat::tests`:

| Test | Asserts |
|------|---------|
| `exec_policy_override_string_allow` | `"allow"` → `ExecSecurityMode::Full` |
| `exec_policy_override_string_deny` | `"deny"` → `ExecSecurityMode::Deny` |
| `exec_policy_override_string_restricted` | `"restricted"` → `ExecSecurityMode::Allowlist` |
| `exec_policy_override_absent_gives_none` | absent field → `None` |
| `exec_policy_override_full_struct_roundtrip` | struct form still works |
| `exec_policy_override_msgpack_none_roundtrip` | old DB blob with no field → `None` (no regression) |

All 279 existing tests pass. Build is clean.

## Verification

Tested against a live v0.3.0 daemon before this PR. After applying the patch and spawning an agent with `exec_policy = "allow"` + `tools = ["shell_exec"]`, the kernel log shows:

```
Agent exec_policy resolved agent=test-v30 exec_mode=Some(Full)
```

And `shell_exec` runs without hitting the approval gate.

### DIFF
--- a/crates/openfang-runtime/src/tool_runner.rs
+++ b/crates/openfang-runtime/src/tool_runner.rs
@@ -131,38 +131,45 @@ pub async fn execute_tool(
         }
     }
 
-    // Approval gate: check if this tool requires human approval before execution
-    if let Some(kh) = kernel {
-        if kh.requires_approval(tool_name) {
-            let agent_id_str = caller_agent_id.unwrap_or("unknown");
-            let input_str = input.to_string();
-            let summary = format!(
-                "{}: {}",
-                tool_name,
-                openfang_types::truncate_str(&input_str, 200)
-            );
-            match kh.request_approval(agent_id_str, tool_name, &summary).await {
-                Ok(true) => {
-                    debug!(tool_name, "Approval granted — proceeding with execution");
-                }
-                Ok(false) => {
-                    warn!(tool_name, "Approval denied — blocking tool execution");
-                    return ToolResult {
-                        tool_use_id: tool_use_id.to_string(),
-                        content: format!(
-                            "Execution denied: '{}' requires human approval and was denied or timed out. The operation was not performed.",
-                            tool_name
-                        ),
-                        is_error: true,
-                    };
-                }
-                Err(e) => {
-                    warn!(tool_name, error = %e, "Approval system error");
-                    return ToolResult {
-                        tool_use_id: tool_use_id.to_string(),
-                        content: format!("Approval system error: {e}"),
-                        is_error: true,
-                    };
+    // Approval gate: check if this tool requires human approval before execution.
+    // Agents with exec_policy mode = Full have pre-approved unrestricted execution;
+    // bypass the gate so that scheduled / headless runs are not blocked waiting for
+    // a human that will never respond.
+    let is_full_exec = exec_policy
+        .is_some_and(|p| p.mode == openfang_types::config::ExecSecurityMode::Full);
+    if !is_full_exec {
+        if let Some(kh) = kernel {
+            if kh.requires_approval(tool_name) {
+                let agent_id_str = caller_agent_id.unwrap_or("unknown");
+                let input_str = input.to_string();
+                let summary = format!(
+                    "{}: {}",
+                    tool_name,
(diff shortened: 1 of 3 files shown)


jaberjaber23:
Fixed in d237ecf. Added exec_policy_lenient deserializer in serde_compat.rs that accepts string shorthand ("allow", "deny", "full", "allowlist") or the full struct table form. The agent.rs exec_policy field now uses this lenient deserializer.

PR #239, opened 2026-03-03
fix: use truncate_str for UTF-8 safe string truncation

slysian:
Replace all `&s[..s.len().min(N)]` patterns with
`openfang_types::truncate_str(s, N)` to prevent panics when the byte index falls inside a multi-byte character (e.g. CJK / emoji).

The bug manifests as a thread panic in session.rs when the thinking content contains non-ASCII text and the 200-byte boundary lands mid-character.

Affected crates: openfang-memory, openfang-kernel, openfang-runtime.


jaberjaber23:
All 7 truncation sites are already fixed on main with truncate_str(). Thanks.

PR #245, opened 2026-03-03
fix: static Linux binaries to eliminate libssl/glibc deps

dyc12389:
## Summary
- **Vendor OpenSSL** in `openfang-channels` for Linux so `native-tls`/`imap` statically link OpenSSL instead of requiring system `libssl.so.3`/`libcrypto.so.3`
- **Add `x86_64-unknown-linux-musl` and `aarch64-unknown-linux-musl`** targets to the release workflow, producing fully static binaries with zero system dependencies
- **Default `install.sh` to musl** on Linux for maximum portability (works on Ubuntu 20.04, older NAS distros, Alpine, etc.)

## Problem
The installer downloads `linux-gnu` binaries built on Ubuntu 22.04, which dynamically link against glibc 2.35 and OpenSSL 3. Systems with older glibc or OpenSSL (e.g. Ubuntu 20.04 NAS) fail at runtime:
```
error while loading shared libraries: libssl.so.3: cannot open shared object file
GLIBC_2.32 not found
GLIBC_2.33 not found
GLIBC_2.34 not found
```

## Test plan
- [x] Verify `cargo build --workspace --lib` passes
- [x] Verify `cargo test --workspace` passes
- [x] Verify musl CI targets build successfully in release workflow
- [x] Install on an Ubuntu 20.04 (or similar older) system and confirm `openfang --version` works
- [x] Verify `ldd` shows `statically linked` (or no dynamic deps) for the musl binary

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -148,6 +148,12 @@ jobs:
           - target: aarch64-unknown-linux-gnu
             os: ubuntu-22.04
             archive: tar.gz
+          - target: x86_64-unknown-linux-musl
+            os: ubuntu-22.04
+            archive: tar.gz
+          - target: aarch64-unknown-linux-musl
+            os: ubuntu-22.04
+            archive: tar.gz
           - target: x86_64-apple-darwin
             os: macos-latest
             archive: tar.gz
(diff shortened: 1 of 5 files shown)


jaberjaber23:
Good addition for static Linux binaries. Has merge conflicts in release.yml. Please rebase onto main.

PR #275, opened 2026-03-03
fix(kernel): fix embedding model config being hardcoded and ignored

woodcoal:
Problem:
- In kernel.rs, embedding driver creation hardcoded the model name
- Config file setting memory.embedding_model was completely ignored
- Always used "text-embedding-3-small" or "nomic-embed-text" instead of user config

Solution:
- Modified create_embedding_driver calls to use config.memory.embedding_model
- Applied config model in all three auto-detection paths:
  1. When embedding_provider is explicitly configured
  2. When auto-detecting OpenAI
  3. When auto-detecting Ollama
- Added logging to show the actual model being used
- Maintained backward compatibility: use defaults when config is empty

Scope:
- crates/openfang-kernel/src/kernel.rs: lines 757-803

Testing:
1. Set custom embedding_model in config file
2. Check logs after startup to verify model name
3. Verify vector search functionality works correctly

### DIFF
--- a/crates/openfang-kernel/src/kernel.rs
+++ b/crates/openfang-kernel/src/kernel.rs
@@ -757,9 +757,14 @@ impl OpenFangKernel {
             if let Some(ref provider) = config.memory.embedding_provider {
                 // Explicit config takes priority
                 let api_key_env = config.memory.embedding_api_key_env.as_deref().unwrap_or("");
-                match create_embedding_driver(provider, "text-embedding-3-small", api_key_env) {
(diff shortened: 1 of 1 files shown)


jaberjaber23:

jaberjaber23:
implemented in v0.3.46 — combined with #413, embedding model config is no longer hardcoded. thanks @woodcoal

jaberjaber23:
Embedding model config is already flexible on main with per-provider defaults and configured_model support. Thanks.

PR #301, opened 2026-03-04
fix(claude-code): support new event-array output format

xinuxZ:
## Summary

  - Fix `claude-code` driver to parse the new event-array JSON output format introduced in Claude Code >= 2.x
  - Maintain backward compatibility with the old single-object `{result, usage}` format

  ## Problem

  Claude Code >= 2.x changed its `--output-format json` output from a single object:

  ```json
  {"result": "...", "usage": {"input_tokens": 10, "output_tokens": 5}}

  to an event array:

  [
    {"type": "system", "session_id": "..."},
    {"type": "assistant", "message": {"content": [{"text": "...", "type": "text"}]}},
    {"type": "result", "result": "...", "usage": {"input_tokens": 100, "output_tokens": 20}}
  ]

  The driver only understood the old format, causing raw JSON to leak through as the agent's response (e.g. into Feishu channels).

  Changes

  File: crates/openfang-runtime/src/drivers/claude_code.rs (+205 -21)

  - complete() — Three-tier parsing with backward compatibility:
    a. Legacy single-object {result, usage} (if result field is present)
    b. New event array [{type: "system"}, {type: "assistant"}, {type: "result"}]
    c. Raw text fallback
  - stream() — Added "assistant" match arm to extract message.content[].text deltas from new format
  - New structs — ClaudeContentPart, ClaudeAssistantMessage, ClaudeEvent for deserializing the event array
  - New function — extract_from_events(): prioritizes result.result over concatenated assistant.message.content[].text; collects usage from result.usage with fallback to assistant.message.usage
  - 5 new tests — Legacy format parsing, event array parsing, fallback when no result field, multi-content-part handling, model flag assertions

  Test plan

  - cargo test -p openfang-runtime -- claude_code — 12 tests pass
  - cargo clippy --workspace --all-targets -- -D warnings — zero warnings
  - Manual: configure provider = "claude-code", send a message, verify clean text response (not raw JSON)

### DIFF
--- a/crates/openfang-runtime/src/drivers/claude_code.rs
+++ b/crates/openfang-runtime/src/drivers/claude_code.rs
@@ -70,9 +70,7 @@ impl ClaudeCodeDriver {
 
     /// Map a model ID like "claude-code/opus" to CLI --model flag value.
     fn model_flag(model: &str) -> Option<String> {
-        let stripped = model
-            .strip_prefix("claude-code/")
-            .unwrap_or(model);
+        let stripped = model.strip_prefix("claude-code/").unwrap_or(model);
         match stripped {
             "opus" => Some("opus".to_string()),
             "sonnet" => Some("sonnet".to_string()),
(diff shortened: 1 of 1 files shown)


jaberjaber23:
Claude Code output format handling was reimplemented in v0.3.25 with fixes for the streaming dead code and legacy format regression issues.

PR #360, opened 2026-03-06
fix(api): correct ToolUse/ToolResult correlation for session display

tuzkier:
## Summary

In the chat UI, tool calls when expanded show empty content. Tool results were not displayed. This PR fixes the serialization logic in `get_agent_session`.

## Root Cause

Per Anthropic protocol, `ToolUse` and `ToolResult` live in **separate messages** (Assistant message and User message respectively). The old code tried to attach `ToolResult` to tools in the **same** message via `tools.last_mut()`, which was always `None` for User messages.

## Changes

- Two-pass approach: build `tool_use_id` → `(msg_idx, tool_idx)` map, then attach `ToolResult` to the correct `ToolUse` in a previous message
- Add `input` field to tool display so expanded view shows call parameters
- Increase result preview from 300 to 2000 chars

## Testing

- [x] Live integration tested: start daemon, send message to agent with tools, verify expanded tool blocks show input and result in dashboard

## Security

- [x] No new unsafe code

## Notes

- Does not affect agent execution; this is a display/serialization bug in `get_agent_session`

### DIFF
--- a/crates/openfang-api/src/routes.rs
+++ b/crates/openfang-api/src/routes.rs
@@ -390,60 +390,84 @@ pub async fn get_agent_session(
 
     match state.kernel.memory.get_session(entry.session_id) {
         Ok(Some(session)) => {
-            let messages: Vec<serde_json::Value> = session
-                .messages
-                .iter()
-                .filter_map(|m| {
-                    let mut tools: Vec<serde_json::Value> = Vec::new();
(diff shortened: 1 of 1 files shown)


jaberjaber23:
The ToolUse/ToolResult correlation fix was implemented in v0.3.25 via the same approach from PR #361.

PR #361, opened 2026-03-06
fix(a2a): A2A protocol compatibility fixes

tuzkier:
## Summary

Fixes A2A (Agent-to-Agent) protocol compatibility issues when connecting to external A2A gateways.

### openfang-runtime (a2a.rs)
- **A2aArtifact**: `name` and `description` changed to `Option<String>` to handle gateways returning `null`
- **A2aTask.status**: Added `A2aTaskStatusWrapper` to support both object form `{"state":"completed"}` and string enum form per A2A spec

### openfang-kernel (kernel.rs)
- **get_a2a_agent_url** / **list_a2a_agents**: Return `card.url` (actual RPC endpoint) instead of discover key (base URL)
- Fixes 404 errors when invoking A2A agents

### openfang-api (routes.rs)
- A2A agents list API now returns `card.url` for proper invocation
- Fix ToolUse/ToolResult cross-message correlation for session display (same fix as #360)

## Testing

- [x] Live integration tested (if applicable): discover A2A agent, invoke task, verify status polling and result display

## Security

- [x] No new unsafe code

### DIFF
--- a/crates/openfang-api/src/routes.rs
+++ b/crates/openfang-api/src/routes.rs
@@ -390,60 +390,84 @@ pub async fn get_agent_session(
 
     match state.kernel.memory.get_session(entry.session_id) {
         Ok(Some(session)) => {
-            let messages: Vec<serde_json::Value> = session
-                .messages
-                .iter()
-                .filter_map(|m| {
-                    let mut tools: Vec<serde_json::Value> = Vec::new();
-                    let content = match &m.content {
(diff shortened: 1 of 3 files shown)


jaberjaber23:
Implemented in v0.3.25. A2A status wrapper, artifact field relaxation, agent URL fix, and ToolUse/ToolResult cross-message correlation all applied.

PR #365, opened 2026-03-06
feat(telegram): support incoming photos, documents, voice and location

### LINKED ISSUE #362: Telegram bridge: incoming photos/media are silently dropped
## Bug Description

Photos, documents, voice messages, and other media sent via Telegram are silently dropped and never reach the agent. Only text messages are processed.

## Root Cause

Two blocking points in the code:

### 1. `crates/openfang-channels/src/telegram.rs` — `parse_telegram_update()`

Line ~480:
```rust
let text = message["text"].as_str()?;
```

This returns `None` for any message without a `text` field (photos, documents, voice, etc.), causing the entire message to be filtered out silently.

### 2. `crates/openfang-channels/src/bridge.rs` — `dispatch_message()`

Line ~456:
```rust
_ => {
    send_response(
        adapter,
        &message.sender,
        "I can only handle text messages for now.".to_string(),
        ...
    )
    .await;
    return;
}
```

Even if media messages were parsed, the bridge explicitly rejects non-text content.

## Infrastructure Already Exists

The type system already supports media:
- `ChannelContent::Image { url, caption }` is defined in `types.rs`
- Outgoing media APIs are fully implemented (`api_send_photo`, `api_send_document`, `api_send_voice`, `api_send_location`)
- Only incoming media parsing and routing is missing

### telegram.rs — `parse_telegram_update()`
1. Check for `message["photo"]` array (Telegram sends multiple sizes, pick the largest)
2. Extract `file_id` from the largest photo
3. Call Telegram Bot API `getFile` to get the file path
4. Construct download URL: `https://api.telegram.org/file/bot{token}/{file_path}`
5. Return `ChannelContent::Image { url, caption }` where caption comes from `message["caption"]`
6. Similarly handle `message["document"]`, `message["voice"]`, `message["location"]`

### bridge.rs — `dispatch_message()`
1. For `ChannelContent::Image`, download the image and either:
   - Pass as base64 data URI to the agent
   - Pass the temporary Telegram URL (valid for ~1 hour)
   - Store locally and pass a local URL
2. Include the caption as context text

## Environment

- OpenFang v0.3.24
- Telegram Bot API
- Claude Code as LLM provider (supports vision)

f-liva:
## Summary

- Photos, documents, voice messages and locations sent via Telegram were silently dropped
- `parse_telegram_update()` required a `text` field — any message without it was filtered out
- The bridge explicitly rejected non-text content with "I can only handle text messages for now"

## Changes

- **telegram.rs**: `parse_telegram_update()` now handles `photo`, `document`, `voice` and `location` message types
- **telegram.rs**: new `telegram_get_file_url()` resolves `file_id` → download URL via `getFile` Bot API
- **bridge.rs**: `Image`/`File`/`Voice`/`Location` content is routed to agents as descriptive text with download URLs
- Function is now `async` to support the `getFile` HTTP call
- All existing tests updated

## Test plan

- [x] Sent a photo via Telegram → agent receives image URL + caption
- [x] Text messages still work as before
- [x] Bot commands still work as before
- [x] `cargo check --package openfang-channels` passes
- [x] All existing tests updated to async and pass

Fixes #362

### DIFF
--- a/crates/openfang-channels/src/bridge.rs
+++ b/crates/openfang-channels/src/bridge.rs
@@ -453,16 +453,22 @@ async fn dispatch_message(
             send_response(adapter, &message.sender, result, thread_id, output_format).await;
             return;
         }
-        _ => {
-            send_response(
-                adapter,
-                &message.sender,
-                "I can only handle text messages for now.".to_string(),
-                thread_id,
-                output_format,
-            )
-            .await;
-            return;
+        ChannelContent::Image { url, caption } => {
+            let cap = caption.as_deref().unwrap_or("");
+            if cap.is_empty() {
+                format!("[User sent an image: {url} ]\nPlease analyze this image.")
+            } else {
+                format!("[User sent an image: {url} ]\nCaption: {cap}")
+            }
+        }
+        ChannelContent::File { url, filename } => {
+            format!("[User sent a file: {filename} — {url} ]")
+        }
+        ChannelContent::Voice { url, duration_seconds } => {
+            format!("[User sent a voice message ({duration_seconds}s): {url} ]")
+        }
+        ChannelContent::Location { lat, lon } => {
+            format!("[User sent a location: lat={lat}, lon={lon}]")
         }
     };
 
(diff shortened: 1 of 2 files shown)


jaberjaber23:
Implemented in v0.3.25. Telegram photo, document, voice, and location media support added with file URL resolution and bridge routing.

PR #377, opened 2026-03-06
fix: detect Claude Code CLI install and auth status

### LINKED ISSUE #376: Claude Code provider card should reflect actual installation and auth status
## Description

The Claude Code provider card in the LLM Providers configuration page has a few UX issues:

### 1. Provider card always shows install instructions and "NO KEY NEEDED"

The Claude Code card always displays the installation command and the label **"NO KEY NEEDED"**, regardless of the actual state. This should be dynamic:

- **Claude Code is installed and authenticated** → show **"CONFIGURED"** instead of "NO KEY NEEDED"
- **Claude Code is installed but NOT authenticated** → show **"NOT SET"** instead of "NO KEY NEEDED"
- **Claude Code is NOT installed** → keep showing the install instructions as today

### 2. Runtime page says Claude Code is not configured even when it is

On the **Runtime** page, when Claude Code is properly installed and authenticated, it still reports that Claude Code is not configured. It should detect the actual status and display accordingly.

### Expected Behavior

The UI should detect whether Claude Code is installed on the system (e.g. `which claude` or checking the expected path) and whether it's authenticated, then reflect the real status in both the Providers page and the Runtime page.

f-liva:
## Summary

- The Claude Code provider card always showed **"No Key Needed"** regardless of actual CLI state, because `detect_auth()` short-circuited all `key_required=false` providers to `NotRequired`
- Now `detect_auth()` handles `claude-code` specifically:
  - **CLI installed + authenticated** → `Configured` → card shows "Configured" (green)
  - **CLI installed, not authenticated** → `Missing` → card shows "Not Set" (warning)
  - **CLI not installed** → `NotRequired` → card shows "No Key Needed" with install instructions
- This also fixes the Runtime/Overview page reporting Claude Code as "not configured" when it's actually ready

Fixes #376

## Test plan

- [ ] Install Claude Code and authenticate → Settings page shows "Configured"
- [ ] Install Claude Code but don't authenticate → Settings page shows "Not Set"
- [ ] No Claude Code installed → Settings page shows "No Key Needed" with install command
- [ ] Runtime/Overview page reflects the correct status

### DIFF
--- a/crates/openfang-runtime/src/model_catalog.rs
+++ b/crates/openfang-runtime/src/model_catalog.rs
@@ -55,6 +55,19 @@ impl ModelCatalog {
     /// Only checks presence — never reads or stores the actual secret.
     pub fn detect_auth(&mut self) {
         for provider in &mut self.providers {
+            // Claude Code: detect CLI installation + authentication
+            if provider.id == "claude-code" {
+                let cli_installed = crate::drivers::claude_code::ClaudeCodeDriver::detect().is_some();
+                if cli_installed && crate::drivers::claude_code::claude_code_available() {
+                    provider.auth_status = AuthStatus::Configured;
+                } else if cli_installed {
+                    provider.auth_status = AuthStatus::Missing;
+                } else {
+                    provider.auth_status = AuthStatus::NotRequired;
+                }
+                continue;
+            }
+
             if !provider.key_required {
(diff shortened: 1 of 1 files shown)


jaberjaber23:
Claude Code auth detection was reimplemented correctly in v0.3.25. The original had a logic bug where the installed-but-not-authenticated state was unreachable.

PR #378, opened 2026-03-06
fix(installer): improve shell detection and fix Fish shell PATH injection

### LINKED ISSUE #372: Installer breaks Fish shell completely by injecting Bash syntax into config.fish
### Description

The OpenFang setup/installation script appends Bash-specific syntax to `~/.config/fish/config.fish` to update the PATH. 

Specifically, it adds:
`export PATH=/home/<user>/.openfang/bin:$PATH`

This syntax is invalid for Fish shell. It destroys the `$PATH` array, leaving the shell unable to find basic system binaries (`uname`, `id`, `ls`, etc.). This causes the shell to break completely upon launch, throwing endless "Unknown command" errors, especially if the user has prompt plugins like `pure` or `done` that rely on standard commands.

### Proposed fix
Instead of using `export PATH=...:$PATH`, the script should inject the standard Fish command for path modification:

```fish
fish_add_path ~/.openfang/bin

### Expected Behavior

The installer should detect if it is modifying a Fish configuration file and use the correct Fish syntax. 

### Steps to Reproduce

1. Use Fish as the default shell.
2. Install/setup OpenFang.
3. Open a new Fish terminal window.
4. See endless `Unknown command: uname` and `Unknown command: id` errors.

### OpenFang Version

0.3.24

### Operating System

Linux (x86_64)

### Logs / Screenshots

_No response_

hobostay:
Fixes #372

## Problem
The installer script only used the `/bin/zsh` environment variable to detect the user's shell. However, `/bin/zsh` reflects the login shell (e.g., /bin/bash or /bin/zsh), not the shell actually being used. When users ran the installer script from Fish shell, it would incorrectly inject Bash export syntax into their `config.fish`, breaking the shell.

## Solution
- Added `detect_shell()` function with multi-layer detection:
  1. Check shell-specific environment variables (`FISH_VERSION`, `ZSH_VERSION`, `BASH`)
  2. Fallback to `/bin/zsh` environment variable
  3. Detect from parent process name via `ps` command
  4. Default to bash for compatibility

- Improved Fish shell handling:
  - Use `fish_add_path` command (recommended for Fish 3.3+)
  - Provide fallback `set -gx PATH` for older Fish versions
  - Avoid injecting Bash syntax into Fish config files

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -111,18 +111,68 @@ install() {
     chmod +x "$INSTALL_DIR/openfang"
 
     # Add to PATH
+    # Detect current shell with fallback to $SHELL, parent process, and defaults
+    detect_shell() {
+        # Check if we're running under fish directly
+        if [ -n "${FISH_VERSION:-}" ]; then
+            echo "fish"
+            return
+        fi
+
+        # Check if we're running under zsh
+        if [ -n "${ZSH_VERSION:-}" ]; then
+            echo "zsh"
+            return
+        fi
+
+        # Check if we're running under bash
+        if [ -n "${BASH:-}" ]; then
+            echo "bash"
+            return
+        fi
+
+        # Fallback to $SHELL environment variable
+        if [ -n "${SHELL:-}" ]; then
+            case "$SHELL" in
+                */fish) echo "fish"; return ;;
+                */zsh) echo "zsh"; return ;;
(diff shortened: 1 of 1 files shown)


jaberjaber23:
Shell detection was reimplemented in v0.3.25. The original had a bug where the BASH env var short-circuited detection, breaking Zsh users.

PR #406, opened 2026-03-07
fix: Browser Hand checks for python instead of python3

### LINKED ISSUE #405: Browser Hand: Python 3 detection shows ✗ even when python3 is installed
## Description

In the Browser Hand activation page, the Python 3 requirement check shows **✗ Python 3 must be installed** even though `python3` is already installed in the system.

Image

Running `apt install python3` from inside the container confirms it's already installed:

```
root@cloud:/# sudo apt install python3
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
python3 is already the newest version (3.11.2-1+b1).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

## Contradictory UI Messages

After clicking through the installation steps, the dashboard shows contradictory messages:

- ✓ Installation complete!
- ✓ `python3` — Python 3 must be installed **installed successfully**
- ✓ `playwright` — Playwright must be installed **is already available**

So it detects python3 and playwright as available, but the initial check still shows ✗.

## Expected Behavior

The Python 3 requirement check should detect that `python3` is already installed and show ✓ from the start, without requiring the user to go through the installation steps.

## Environment

- OpenFang v0.3.26 running in Docker (debian:bookworm-slim)
- `python3` installed via apt
- `playwright` installed via pip3
- Chromium installed via `playwright install --with-deps chromium`

f-liva:
## Summary

- The Browser Hand `HAND.toml` sets `check_value = "python"` for the Python 3 requirement check
- On Debian/Ubuntu, the binary is `python3` — there is no `python` symlink by default
- This causes the requirement check to show ✗ even when Python 3 is installed
- Changed `check_value` to `"python3"` to match the actual binary name

Fixes #405

## Test plan

- [x] Deploy on Debian-based system with `python3` installed (no `python` symlink)
- [x] Browser Hand requirement check should show ✓ for Python 3
- [x] Verify Playwright check still works

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2507,15 +2507,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "html-escape"
-version = "0.2.13"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6d1ad449764d627e22bfd7cd5e8868264fc9236e07c752972b4080cd351cb476"
-dependencies = [
- "utf8-width",
-]
-
-[[package]]
 name = "html5ever"
 version = "0.29.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3875,7 +3866,7 @@ dependencies = [
 
 [[package]]
 name = "openfang-api"
-version = "0.3.25"
+version = "0.3.24"
 dependencies = [
  "async-trait",
  "axum",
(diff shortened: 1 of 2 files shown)


jaberjaber23:
This adds a custom CI workflow and version tracking file that don't match the project's patterns. Closing.

PR #410, opened 2026-03-07
Fix potential panic bugs in string slicing and server shutdown

hobostay:
## Summary

This PR fixes several potential panic bugs and race conditions found in the codebase:

### 1. Race Condition in Desktop Server Shutdown (`crates/openfang-desktop/src/server.rs`)

**Problem**: Both `shutdown()` and `Drop` implementation tried to send shutdown signals and join threads, which could lead to race conditions when both try to manipulate the same resources simultaneously.

**Fix**: 
- Added `AtomicBool` flag (`shutdown_initiated`) to track shutdown state
- Both `shutdown()` and `Drop` now use `compare_exchange` to ensure shutdown only happens once
- This prevents duplicate shutdown attempts and potential race conditions

### 2. String Slicing Panic (`crates/openfang-api/src/channel_bridge.rs`)

**Problem**: Multiple locations used `&str[..8]` to truncate UUID strings for display without checking if the string was at least 8 characters long. While UUIDs are always 36 characters, this pattern is unsafe and could cause panics if the data changes.

**Fix**:
- Added `truncate_str()` helper function that safely returns the full string if it's shorter than the requested length
- Fixed 9 occurrences at lines: 354, 393, 408, 431, 491, 513, 545, 565, 609

### 3. String Slicing Panic (`crates/openfang-runtime/src/tool_runner.rs`)

**Problem**: Canvas filename generation used `&canvas_id[..8]` without checking length.

**Fix**: 
- Added length check before slicing
- Falls back to full string if shorter than 8 characters

## Test Plan

- [x] Build passes: `cargo build --workspace`
- [x] Tests pass: `cargo test --workspace`
- [x] No new clippy warnings introduced
- [x] Manual code review of all changes

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

### DIFF
--- a/crates/openfang-api/src/channel_bridge.rs
+++ b/crates/openfang-api/src/channel_bridge.rs
@@ -56,6 +56,16 @@ use std::sync::Arc;
 use std::time::{Duration, Instant};
 use tracing::{error, info, warn};
 
+/// Safely truncate a string to at most n characters.
+/// Returns the original string if it's shorter than n characters.
+fn truncate_str(s: &str, n: usize) -> &str {
+    if s.len() > n {
+        &s[..n]
+    } else {
+        s
+    }
+}
+
 /// Wraps `OpenFangKernel` to implement `ChannelBridgeHandle`.
 pub struct KernelBridgeAdapter {
     kernel: Arc<OpenFangKernel>,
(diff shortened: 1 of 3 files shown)


jaberjaber23:

jaberjaber23:
implemented in v0.3.46 — fixed 9 string slicing instances in channel_bridge using our safe_truncate_str + desktop shutdown race. some overlapped with our v0.3.43 fix. thanks @hobostay

PR #417, opened 2026-03-07
fix: default hourly cost quota should be unlimited (0.0), not $1.00

### LINKED ISSUE #416: Default ResourceQuota enforces $1/hour cost limit even when no quota is configured
## Bug Description

When no cost quota is explicitly configured for an agent, the system defaults to a **$1.00/hour** limit, causing agents to be throttled unexpectedly.

## Root Cause

In `crates/openfang-types/src/agent.rs`, `ResourceQuota::default()` sets:

```rust
max_cost_per_hour_usd: 1.0,   // ← hardcoded $1/hour limit
max_cost_per_day_usd: 0.0,    // unlimited
max_cost_per_month_usd: 0.0,  // unlimited
```

The hourly cost quota defaults to `1.0`, while daily and monthly default to `0.0` (unlimited). This is inconsistent — a user who hasn't configured any quota expects **no limits**, not a hidden $1/hour cap.

## Observed Behavior

Error returned to users after normal usage:

```
Agent error: Resource quota exceeded: Agent 2ab6a9bf-ffb8-48df-86e5-9797e5e13c01 exceeded hourly cost quota: $1.0097 / $1.0000
```

The user had **no quota configured** — the $1.00 limit comes entirely from `ResourceQuota::default()`.

## Expected Behavior

`max_cost_per_hour_usd` should default to `0.0` (unlimited), consistent with the daily and monthly defaults. The metering check in `crates/openfang-kernel/src/metering.rs` already handles `0.0` correctly:

```rust
if quota.max_cost_per_hour_usd > 0.0 {
    // only enforced when > 0
}
```

## Suggested Fix

```rust
impl Default for ResourceQuota {
    fn default() -> Self {
        Self {
            // ...
            max_cost_per_hour_usd: 0.0,  // unlimited by default (was 1.0)
            max_cost_per_day_usd: 0.0,   // unlimited
            max_cost_per_month_usd: 0.0, // unlimited
        }
    }
}
```

## Impact

Any agent using `ResourceQuota::default()` (which includes `wizard.rs`, `registry.rs`, `kernel.rs`) hits this invisible ceiling after ~$1 of usage per hour — especially problematic for agents using expensive models like Claude Opus.

## Screenshots

Image
Image

f-liva:
## Summary

- Changed `ResourceQuota::default()` `max_cost_per_hour_usd` from `1.0` to `0.0` (unlimited), consistent with daily and monthly defaults
- Fixed `apply_budget_defaults()` in `kernel.rs` which compared against the old hardcoded `1.0` sentinel value

## Problem

When no quota is configured, `max_cost_per_hour_usd` defaults to `1.0` while `max_cost_per_day_usd` and `max_cost_per_month_usd` default to `0.0` (unlimited). This creates a hidden $1/hour cap that blocks agents unexpectedly.

## Test plan

- [x] Existing test `test_check_quota_zero_limit_skipped` validates that `0.0` means no enforcement
- [x] Tests `test_record_and_check_quota_under` and `test_check_quota_exceeded` set explicit values, unaffected

Fixes #416

### DIFF
--- a/crates/openfang-kernel/src/kernel.rs
+++ b/crates/openfang-kernel/src/kernel.rs
@@ -4656,8 +4656,8 @@ fn apply_budget_defaults(
     budget: &openfang_types::config::BudgetConfig,
     resources: &mut ResourceQuota,
 ) {
-    // Only override hourly if agent has the built-in default (1.0) and global is set
-    if budget.max_hourly_usd > 0.0 && resources.max_cost_per_hour_usd == 1.0 {
+    // Only override hourly if agent has unlimited default (0.0) and global is set
+    if budget.max_hourly_usd > 0.0 && resources.max_cost_per_hour_usd == 0.0 {
         resources.max_cost_per_hour_usd = budget.max_hourly_usd;
     }
     // Only override daily/monthly if agent has unlimited (0.0) and global is set
--- a/crates/openfang-types/src/agent.rs
+++ b/crates/openfang-types/src/agent.rs
@@ -267,7 +267,7 @@ impl Default for ResourceQuota {
             max_tool_calls_per_minute: 60,
             max_llm_tokens_per_hour: 1_000_000,
             max_network_bytes_per_hour: 100 * 1024 * 1024, // 100 MB
-            max_cost_per_hour_usd: 1.0,
+            max_cost_per_hour_usd: 0.0,   // unlimited
             max_cost_per_day_usd: 0.0,   // unlimited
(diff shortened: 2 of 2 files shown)


jaberjaber23:

jaberjaber23:
implemented in v0.3.46 — default token quota is now unlimited (0). thanks @f-liva

jaberjaber23:
Already fixed on main.

PR #424, opened 2026-03-08
Stop agent follow-up actions after tool failures

### LINKED ISSUE #414: [BUG] Agent continues generating fabricated response and calling tools after web_search failure
### Description

 When using researcher-hand, if web_search fails (e.g., no results returned, DuckDuckGo blocked), the LLM does not stop. Instead, it:

  1. Fabricates search results and outputs a conclusion as if the search succeeded
  2. Continues calling downstream tools (file_write, memory_store) based on the fabricated data

### Expected Behavior

 Agent reports the search failure to the user and stops
 Actual: Agent fabricates an analysis, writes it to a file via file_write, and stores it in memory via memory_store

### Steps to Reproduce

 1. Start an agent using researcher-hand with DeepSeek as the LLM provider
  2. Ask the agent to research a topic (e.g., "Research silver price forecast 2026")
  3. web_search tool fails — returns no results
  4. Expected: Agent reports the search failure to the user and stops
  5. Actual: Agent fabricates an analysis, writes it to a file via file_write, and stores it in memory via memory_store

### OpenFang Version

0.3.26

### Operating System

Windows

### Logs / Screenshots

Image
 Environment

  - OS: Windows 11
  - Provider: DeepSeek (deepseek-chat)
  - Version: 0.3.26

tsubasakong:
## Summary

Fixes #414

- inject explicit no-fabrication guidance into the next LLM turn whenever any tool call fails
- apply the guidance in both standard and streaming agent loops so failed `web_search` results are not treated as evidence for downstream actions
- add a regression test that asserts the guidance is included after a failed tool call

## Changes

- inject explicit no-fabrication guidance into the next LLM turn whenever any tool call fails
- apply the guidance in both standard and streaming agent loops so failed `web_search` results are not treated as evidence for downstream actions
- add a regression test that asserts the guidance is included after a failed tool call

## Testing

Testing notes:
- `git diff --check`
- host Rust toolchain unavailable here (`cargo`/`rustc` not installed), so I could not run `cargo test`

Fixes #414.

## Security

- [x] No new unsafe code

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3792,7 +3792,7 @@ dependencies = [
 
 [[package]]
 name = "openfang-api"
-version = "0.3.47"
+version = "0.4.0"
 dependencies = [
  "async-trait",
  "axum",
@@ -3801,6 +3801,8 @@ dependencies = [
  "dashmap",
  "futures",
  "governor",
+ "hex",
+ "hmac",
  "openfang-channels",
  "openfang-extensions",
  "openfang-hands",
@@ -3814,6 +3816,7 @@ dependencies = [
  "reqwest 0.12.28",
  "serde",
  "serde_json",
+ "sha2",
  "socket2 0.5.10",
  "subtle",
  "tempfile",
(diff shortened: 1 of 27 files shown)


jaberjaber23:
thanks for this! we're reviewing and may implement this differently to fit our architecture. will keep you posted

PR #433, opened 2026-03-08
fix: set default User-Agent on OpenAI driver HTTP client

ozekimasaki:
## Summary

- `reqwest::Client::new()` does not set a `User-Agent` header by default
- Some OpenAI-compatible API providers and proxies reject requests without `User-Agent`, returning **405 errors**
- This sets `OpenFang/1.0` as a sensible default, with a graceful fallback to `reqwest::Client::new()` if the builder fails

## Changes

**File:** `crates/openfang-runtime/src/drivers/openai.rs` (1 file, +4 −1)

```rust
// Before
client: reqwest::Client::new(),

// After
client: reqwest::Client::builder()
    .user_agent("OpenFang/1.0")
    .build()
    .unwrap_or_else(|_| reqwest::Client::new()),
```

## Test plan

- [ ] Existing tests pass (no tests assert on User-Agent)
- [ ] Non-breaking, backward compatible — only adds a header that was previously absent
- [ ] Verified against providers that require User-Agent

### DIFF
--- a/crates/openfang-runtime/src/drivers/openai.rs
+++ b/crates/openfang-runtime/src/drivers/openai.rs
@@ -25,7 +25,10 @@ impl OpenAIDriver {
         Self {
             api_key: Zeroizing::new(api_key),
             base_url,
-            client: reqwest::Client::new(),
+            client: reqwest::Client::builder()
+                .user_agent("OpenFang/1.0")
+                .build()
+                .unwrap_or_else(|_| reqwest::Client::new()),
             extra_headers: Vec::new(),
         }
     }


zxl-coder-ai:
The openai adapter already has a with_extra_headers method, it's just not being called during registration. Wouldn't it be more elegant if your PR supported passing the User-Agent via configuration instead?

ozekimasaki:
Thanks for the feedback!

I considered that, but I think a sensible default User-Agent belongs at the client level rather than in configuration:

- **Every HTTP client should identify itself** — this is basic HTTP hygiene, not a user-facing setting. `reqwest::Client::new()` sending no User-Agent at all is the gap this fixes.
- **Configuration adds friction** — if User-Agent requires config, users who don't set it will hit the same 405 errors silently. A default that "just works" is safer.
- **`with_extra_headers` serves a different purpose** — it's designed for per-deployment auth headers (e.g. Copilot IDE tokens), not for setting a baseline client identity.

The two aren't mutually exclusive — users can still override User-Agent via `with_extra_headers` if needed. This PR just ensures there's always a reasonable default in place.

jaberjaber23:
implemented in v0.3.46 — added User-Agent header to all 3 LLM drivers + web fetch client. thanks @ozekimasaki

jaberjaber23:
Already on main.

PR #438, opened 2026-03-08
fix(chat-ui): prevent Enter send during IME composition

### LINKED ISSUE #429: Chat UI: Enter key sends message during CJK IME composition
## Bug Description

When using Chinese (CJK) Input Method Editor (IME) in the web chat UI, pressing Enter to confirm/select a character from the candidate list immediately sends the message instead of just confirming the character selection.

## Steps to Reproduce

1. Open the OpenFang dashboard chat (`http://127.0.0.1:4200`)
2. Switch to a Chinese IME (e.g., Pinyin)
3. Type pinyin characters (e.g., "ni hao")
4. When the candidate list appears, press Enter to select a character
5. **Expected**: The selected character is inserted into the input field
6. **Actual**: The incomplete/partial text is sent as a message immediately

## Expected Behavior

Pressing Enter during IME composition should only confirm the character selection. The message should only be sent when Enter is pressed **after** composition ends (i.e., when `isComposing` is `false`).

## Suggested Fix

Add an `isComposing` check to the keydown event handler:

```javascript
textarea.addEventListener('keydown', (e) => {
  if (e.key === 'Enter' && !e.isComposing) {
    sendMessage();
  }
});
```

## Environment

- OS: macOS (Apple Silicon)
- OpenFang version: 0.3.29
- IME: macOS built-in Chinese Pinyin

## References

- Same issue in Open WebUI: https://github.com/open-webui/open-webui/issues/16608
- Same issue in elizaOS: https://github.com/elizaOS/eliza/issues/2272

pandego:
## Summary
Fix chat input Enter handling for CJK IME composition in the dashboard UI.

When users are composing text with an IME, pressing Enter should confirm the current candidate, not send the message.

## What changed
- Updated the chat textarea Enter handler in `crates/openfang-api/static/index_body.html`
- Added `!$event.isComposing` guard before calling `sendMessage()`

## Issue
Closes #429

## Validation
- Manual logic validation against browser IME behavior:
  - Enter sends only when `isComposing` is false
  - Shift+Enter behavior remains unchanged
- Local Rust checks were not runnable in this environment because `cargo` is not installed.

### DIFF
--- a/crates/openfang-api/static/index_body.html
+++ b/crates/openfang-api/static/index_body.html
@@ -735,7 +735,7 @@
                   <span class="text-xs" style="color:var(--danger)" x-text="formatRecordingTime()"></span>
                 </div>
                 <textarea id="msg-input" rows="1" :placeholder="recording ? 'Recording... release to send' : 'Message OpenFang... (/ for commands)'"
-                          @keydown.enter.prevent="if(!$event.shiftKey){if(showModelPicker && filteredModelPicker.length){pickModel(filteredModelPicker[modelPickerIdx].id)}else if(showSlashMenu && filteredSlashCommands.length){executeSlashCommand(filteredSlashCommands[slashIdx].cmd)}else{sendMessage()}}"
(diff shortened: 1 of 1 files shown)


jaberjaber23:
implemented in v0.3.46 — we added IME composition guards to all 6 input fields in the dashboard. thanks for catching this @pandego

pandego:
Thanks for the update. Since the IME composition guard landed upstream in v0.3.46, I am closing this PR as superseded to avoid duplicate noise.

Appreciate the quick follow-through on the fix.

PR #453, opened 2026-03-08
Fix CSP: add 'unsafe-eval' to script-src for dashboard

citadelgrad:
## Summary

- Adds `'unsafe-eval'` to the `script-src` directive in the Content-Security-Policy header
- The dashboard's inline JS framework (Alpine.js/Petite-Vue) uses `new AsyncFunction()`, which requires `'unsafe-eval'` — without it the browser blocks execution and throws dozens of `EvalError` console errors

## Changed file

`crates/openfang-api/src/middleware.rs` — one-line change in `security_headers()` middleware

## Verification

```bash
curl -sI http://127.0.0.1:4200/ | grep -i content-security-policy
```

Should now include `script-src 'self' 'unsafe-inline' 'unsafe-eval'` and the dashboard should load without `EvalError` console errors.

### DIFF
--- a/crates/openfang-api/src/middleware.rs
+++ b/crates/openfang-api/src/middleware.rs
@@ -202,7 +202,7 @@ pub async fn security_headers(request: Request<Body>, next: Next) -> Response<Bo
     // All JS/CSS is bundled inline — only external resource is Google Fonts.
     headers.insert(
         "content-security-policy",
-        "default-src 'self'; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com https://fonts.gstatic.com; img-src 'self' data: blob:; connect-src 'self' ws://localhost:* ws://127.0.0.1:* wss://localhost:* wss://127.0.0.1:*; font-src 'self' https://fonts.gstatic.com; media-src 'self' blob:; frame-src 'self' blob:; object-src 'none'; base-uri 'self'; form-action 'self'"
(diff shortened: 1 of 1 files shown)


citadelgrad:
## Security Review: `'unsafe-eval'` CSP Addition

Performed a focused security review of this change. **No exploitable vulnerabilities introduced by this PR.**

### Analysis Summary

Four potential findings were investigated and all filtered as false positives:

| Finding | Concern | Verdict |
|---|---|---|
| XSS via Alpine.js directive injection | Attacker injects `x-*` attributes through `x-html` | **False positive** — Alpine.js does NOT compile directives on HTML injected via `x-html`; it only sets `innerHTML` |
| marked.js + `x-html` rendering | Unsanitized Markdown → HTML via `x-html` | **Pre-existing** — `'unsafe-inline'` already permits inline event handlers (`onclick`, `onerror`); unrelated to this PR |
| Chat message flow | LLM responses rendered without sanitization | **Pre-existing** — same root cause as above |
| Weak CSP | `'unsafe-inline'` + `'unsafe-eval'` = minimal CSP protection | **Architectural concern**, not an exploitable vuln introduced here |

### Why `'unsafe-eval'` is needed

Alpine.js 3.15.8 uses `new Function()` internally to compile expressions in `x-show`, `@click`, `:class`, etc. Without `'unsafe-eval'`, the browser throws `EvalError` and the dashboard JS breaks.

### Recommendations for follow-up

These are **not blockers** for this PR but worth tracking:

1. **Add DOMPurify** to sanitize `renderMarkdown()` output before `x-html` injection (addresses the pre-existing unsanitized Markdown pipeline)
2. **Long-term**: Migrate to nonce-based CSP with precompiled Alpine templates to eliminate both `'unsafe-inline'` and `'unsafe-eval'`

PR #467, opened 2026-03-09
fix: correct Slack adapter message fields for RBAC and DM policy

zilvinasu:
## Summary

The Slack adapter has four related issues that break RBAC user matching, DM policy enforcement, and DM delivery:

1. **`platform_id` set to wrong value for RBAC** - `platform_id` is used by the send path (`chat.postMessage` needs a channel ID), and all other adapters set it to channel/chat ID. But RBAC's `identify()` matches `channel_bindings` against `platform_id`, expecting a user ID. Fix: store the Slack user ID in `metadata["sender_user_id"]` and add a `sender_user_id()` helper in `bridge.rs` that reads it for RBAC and rate limiting.

2. **`is_group` hardcoded to `true`** - DMs (channels starting with `D`) were classified as group messages, so `dm_policy` (ignore/allowed_only/respond) was never enforced. Fix: `is_group = !channel.starts_with('D')`.

3. **Bridge RBAC reads `platform_id` instead of user identity** - `authorize_channel_user()` and `rate_limiter.check()` used `message.sender.platform_id`, which is the channel ID. Fix: use `sender_user_id()` helper that checks metadata first, falling back to `platform_id` for adapters that don't set `sender_user_id`.

4. **`allowed_channels` filter blocks DMs** - The `allowed_channels` whitelist applies to all messages including DMs. Since DM channel IDs (`D`-prefixed) are never in the allowed list, all DMs are silently dropped before `dm_policy` can evaluate them. Fix: exempt `D`-prefixed channels from the `allowed_channels` filter — DM access control is handled by `dm_policy`.

## Changes

- \`crates/openfang-channels/src/slack.rs\` - store user ID in metadata, revert platform_id to channel, detect DMs via channel prefix, exempt DMs from allowed_channels filter
- \`crates/openfang-channels/src/bridge.rs\` - add \`sender_user_id()\` helper, use it for RBAC and rate limiting

## Backward compatibility

- The \`sender_user_id()\` helper falls back to \`platform_id\` when metadata key is absent, so all other adapters continue working unchanged
- \`platform_id\` stays as channel ID (matching every other adapter), so the send path is unaffected
- DMs were previously always blocked by \`allowed_channels\`, so exempting them is strictly additive

## Checklist

- [x] \`cargo fmt\` - our files are clean (workspace has pre-existing \`ar_archive_writer\` build issue on stable)
- [x] Tests updated: existing assertions corrected, new test for DM detection added
- [x] No new \`unsafe\` code

### DIFF
--- a/crates/openfang-channels/src/slack.rs
+++ b/crates/openfang-channels/src/slack.rs
@@ -370,8 +370,11 @@ async fn parse_slack_event(
 
     let channel = event["channel"].as_str()?;
 
-    // Filter by allowed channels
-    if !allowed_channels.is_empty() && !allowed_channels.contains(&channel.to_string()) {
+    // Filter by allowed channels (DMs are exempt — dm_policy handles those)
+    if !channel.starts_with('D')
+        && !allowed_channels.is_empty()
+        && !allowed_channels.contains(&channel.to_string())
+    {
         return None;
     }
 


zilvinasu:
@jaberjaber23 would appreciate a review when you get a chance. This fixes three related issues in the Slack adapter that break RBAC user matching and DM policy enforcement.

zilvinasu:
btw, tested on the deployed openfang, found another bug for allowed_channels check, as it blocks DMs by default as well, will probably fix it as part of this

zilvinasu:
fixed now, DM RBAC works fine, manual testing done as well, let me know if anything missing @jaberjaber23

jaberjaber23:
thanks for this! we're reviewing and may implement this differently to fit our architecture. will keep you posted

jaberjaber23:
Incorporated sender_user_id() helper into main for correct Slack RBAC. Thanks.

PR #489, opened 2026-03-10
Fix ClawHub install rate limiting by preferring /file endpoint

mccar2cm:
## Summary
- ClawHub's `/download` endpoint has a 20 req/window rate limit shared across IPs, causing skill installs to fail with 429 even on the first attempt (common on shared hosting/VPS)
- The `/file` endpoint has a 120 req/window limit and returns SKILL.md content directly — sufficient for the majority of ClawHub skills
- Now tries `/skills/{slug}/file?path=SKILL.md` first, falls back to `/download` only for non-SKILL.md packages (zips, Node.js)
- Respects the `retry-after` header on 429 responses instead of fixed exponential backoff
- Dashboard now shows the actual server error message on 500s instead of the generic "Server error — check daemon logs"

## Test plan
- [ ] Install a SKILL.md-based skill from ClawHub (e.g., `agentmail`) — should succeed via `/file` endpoint
- [ ] Install a non-SKILL.md skill (zip/package) — should fall back to `/download`
- [ ] Verify `cargo test --workspace` passes (1886 tests)
- [ ] Verify `cargo clippy --workspace --all-targets -- -D warnings` is clean
- [ ] Trigger a 500 error in dashboard and verify the real error message is shown

### DIFF
--- a/crates/openfang-api/static/js/api.js
+++ b/crates/openfang-api/static/js/api.js
@@ -123,7 +123,7 @@ function friendlyError(status, serverMsg) {
   if (status === 404) return serverMsg || 'Resource not found';
   if (status === 429) return 'Rate limited — slow down and try again';
   if (status === 413) return 'Request too large';
-  if (status === 500) return 'Server error — check daemon logs';
+  if (status === 500) return serverMsg || 'Server error — check daemon logs';
   if (status === 502 || status === 503) return 'Daemon unavailable — is it running?';
   return serverMsg || 'Unexpected error (' + status + ')';
 }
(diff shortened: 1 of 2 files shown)


Jengro777:
#492

jaberjaber23:
thanks for this! we already implemented ClawHub rate limit retry with 5-attempt exponential backoff + jitter in v0.3.44 (fix for #492). appreciate the contribution

Jengro777:
> thanks for this! we already implemented ClawHub rate limit retry with 5-attempt exponential backoff + jitter in v0.3.44 (fix for #492). appreciate the contribution


Is it because the project is still in the early stages? Would it be better to accept community contributions?

mccar2cm:
No worries - I liked the idea of hitting the files endpoint then the download. Take a look, see if you think that is an interesting approach.

Jengro777:
image

jaberjaber23:
ClawHub /file endpoint is already on main. Thanks.

PR #497, opened 2026-03-10
fix: concurrent message dispatch to prevent voice message loss

### LINKED ISSUE #495: [Bug] Voice messages lost when multiple sent in quick succession (sequential dispatch)
## Problem

When a user sends multiple voice messages (or any messages) in quick succession via Telegram, only the **last** message's response reaches the user in a timely manner. Earlier messages are effectively "lost" from the user's perspective.

## Root Cause

In `bridge.rs` line 309, `dispatch_message()` is **awaited sequentially** inside the stream loop:

```rust
msg = stream.next() => {
    match msg {
        Some(message) => {
            dispatch_message(  // <-- blocks the loop
                &message, &handle, &router,
                adapter_clone.as_ref(), &rate_limiter,
            ).await;           // <-- until LLM responds (10-30s+)
        }
```

Since `dispatch_message` calls `send_message` which invokes an LLM (10-30+ seconds), the stream loop is **blocked** while waiting for the LLM response. Messages 2 and 3 sit in the mpsc buffer and are only processed after message 1's LLM call completes.

For CLI agent sessions forwarded via the bridge, this means the session receives message 1, processes it (including LLM response), and by the time messages 2 and 3 arrive, the session context may have moved on — making earlier messages appear "lost."

## Steps to Reproduce

1. Configure a Telegram channel with an agent
2. Send 3 voice messages in quick succession (within ~5 seconds)
3. Only the last voice message gets a timely response; the first two are either delayed significantly or their responses arrive out of context

## Expected Behavior

All messages should be dispatched concurrently (or at least not block the polling loop). Each message should be processed independently so that rapid-fire messages don't queue behind slow LLM calls.

## Proposed Fix

Spawn `dispatch_message` as a concurrent task instead of awaiting it inline:

```rust
Some(message) => {
    let handle = handle.clone();
    let router = router.clone();
    let adapter = adapter_clone.clone();
    let rate_limiter = rate_limiter.clone();
    tokio::spawn(async move {
        dispatch_message(
            &message, &handle, &router,
            adapter.as_ref(), &rate_limiter,
        ).await;
    });
}
```

This preserves message ordering at the Telegram polling level while allowing concurrent dispatch. An optional semaphore could limit max concurrent dispatches per channel if needed.

## Environment

- OpenFang v0.3.34
- Telegram channel adapter with long-polling
- Agent running in CLI session (Claude Code)

## Related

- #375 — Voice messages not forwarded to CLI sessions (separate routing issue)

f-liva:
## Summary

- **Spawns `dispatch_message` as a concurrent task** instead of awaiting it sequentially in the stream loop
- Prevents messages from being lost when multiple arrive in quick succession (e.g. 3 voice messages within seconds)
- Each dispatch runs independently so slow LLM calls (10-30s+) don't block the polling loop

## Problem

`dispatch_message()` in `bridge.rs` was awaited inline, blocking the stream loop. When a user sends 3 voice messages rapidly, message 1 triggers an LLM call that takes ~20s. Messages 2 and 3 sit in the mpsc buffer and are only processed after the LLM finishes — making them appear "lost" from the user's perspective.

## Changes

- `crates/openfang-channels/src/bridge.rs`: Clone `Arc` handles and spawn dispatch as a `tokio::spawn` task

## Test plan

- [ ] `cargo build --workspace --lib` compiles
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [ ] Manual test: send 3 voice messages in quick succession via Telegram, verify all 3 get responses

Fixes #495

### DIFF
--- a/crates/openfang-channels/src/bridge.rs
+++ b/crates/openfang-channels/src/bridge.rs
@@ -306,13 +306,23 @@ impl BridgeManager {
                     msg = stream.next() => {
                         match msg {
                             Some(message) => {
-                                dispatch_message(
-                                    &message,
-                                    &handle,
-                                    &router,
-                                    adapter_clone.as_ref(),
-                                    &rate_limiter,
-                                ).await;
+                                // Spawn each dispatch concurrently so that slow LLM calls
+                                // don't block subsequent messages in the stream.
+                                // Fixes: multiple messages sent in quick succession being
+                                // effectively lost because dispatch was sequential.
+                                let h = handle.clone();
+                                let r = router.clone();
+                                let a = adapter_clone.clone();
+                                let rl = rate_limiter.clone();
+                                tokio::spawn(async move {
+                                    dispatch_message(
+                                        &message,
+                                        &h,
+                                        &r,
+                                        a.as_ref(),
+                                        &rl,
+                                    ).await;
+                                });
                             }
                             None => {
                                 info!("Channel adapter {} stream ended", adapter_clone.name());


f-liva:
## Build Verification ✅

Compilation tested successfully in the container environment:

- `cargo check -p openfang-channels` — **passed** (crate containing the fix)
- `cargo check --workspace --exclude openfang-desktop` — **passed** (full workspace minus GTK-dependent crate)
- Rust toolchain: **1.75.0** (project minimum)

`openfang-desktop` excluded because the container lacks GTK/GLib dev libraries — unrelated to this change.

🤖 Tested by Ambrogio

jaberjaber23:
thanks for this! we already implemented concurrent message dispatch with a Semaphore(32) in v0.3.44 (fix for #495). your approach was similar — appreciate the contribution

jaberjaber23:
Already on main with semaphore-bounded concurrency (32 tasks). Thanks for identifying the issue.

PR #518, opened 2026-03-11
feat: full Claude Code OAuth provider support

blackopsrepl:
## What this PR does

Adds full Anthropic support **beyond the direct API key** — specifically, users who authenticate via Claude Code (OAuth) can now use their existing Claude Code session as the LLM backend, without needing a separate `ANTHROPIC_API_KEY`.

Previously the only way to use Anthropic models was a raw API key. This PR adds a second path: if you have Claude Code installed and authenticated, OpenFang can delegate all LLM calls to the `claude` CLI subprocess, which handles auth internally.

## How it works

When the user selects Anthropic in the init wizard and chooses "Claude Code OAuth", the wizard now writes `provider = "claude-code"` to config instead of `provider = "anthropic"`. This routes all requests through `ClaudeCodeDriver` (which shells out to `claude -p`) rather than hitting `api.anthropic.com` directly with a Bearer token — which Anthropic's API rejects.

## Changes

- **`init_wizard.rs`** — OAuth auth method picker (API key vs Claude Code OAuth), `use_claude_code_oauth` state flag, correct provider written to config, auto-detection of existing Claude Code credentials
- **`drivers/mod.rs`** — remove the OAuth token fallback from the `anthropic` branch (it never worked; `AnthropicDriver` only supports API keys)
- **`drivers/anthropic.rs`** — remove dead `AuthMode::OAuth` / `new_oauth()` / Bearer header code
- **`drivers/claude_code.rs`** — three runtime fixes required to make the driver actually work:
  - Unset `CLAUDECODE` env var before spawning (CLI refuses to run in nested sessions)
  - Fix stream parser: current CLI emits `message.content[{type,text}]` blocks, not a flat string
  - Fix `complete()` parser: same nested block-array format for JSON output

## Test plan

- [ ] `openfang init` → Anthropic → Claude Code OAuth → model list shows claude-code models
- [ ] `~/.openfang/config.toml` has `provider = "claude-code"`
- [ ] Messages return responses (no "OAuth not supported" error)
- [ ] Anthropic with `ANTHROPIC_API_KEY` still works normally

### DIFF
--- a/crates/openfang-cli/src/tui/screens/init_wizard.rs
+++ b/crates/openfang-cli/src/tui/screens/init_wizard.rs
@@ -280,6 +280,19 @@ enum KeyTestState {
     Warn,
 }
 
+/// Sub-state for the Anthropic auth method picker (shown before key entry).
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum AnthropicAuthMode {
+    /// Show "API key / Claude Code OAuth" choice list.
+    Pick,
+    /// User chose API key — show the key entry UI.
+    EnterKey,
+    /// User chose OAuth and the credentials file is present.
+    OAuthReady,
+    /// User chose OAuth but the credentials file is missing.
+    OAuthMissing,
+}
+
 /// A model entry for list display.
 struct ModelEntry {
     id: String,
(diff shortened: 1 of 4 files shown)


blackopsrepl:
Doesn't work with tools. Switching to draft until further investigation.

jaberjaber23:
Looks good but this is still a draft. Mark as ready when you're done and we'll review.

jaberjaber23:

PR #529, opened 2026-03-11
fix(channels): detect image media type from magic bytes

### LINKED ISSUE #528: bug(channels): Telegram image media_type is application/octet-stream — breaks vision
## Bug Description

Photos sent via Telegram are downloaded and base64-encoded correctly, but the `media_type` is set to `application/octet-stream` instead of the actual image type (e.g. `image/jpeg`). This causes LLM providers to reject the image content, as they require a valid image MIME type.

## Regression

This was supposed to be fixed as part of issue #362 and the related PR #365. The Telegram parsing and bridge download pipeline was added, but the media type detection has a bug that silently breaks vision.

## Root Cause

In `crates/openfang-channels/src/bridge.rs`, `download_image_to_blocks()` (line ~818):

```rust
let content_type = resp
    .headers()
    .get("content-type")
    .and_then(|v| v.to_str().ok())
    .map(|ct| ct.split(';').next().unwrap_or(ct).trim().to_string());

let media_type = content_type.unwrap_or_else(|| {
    // fallback to URL extension detection
    ...
});
```

The code trusts the `Content-Type` header from Telegram's file download API. However, **Telegram returns `application/octet-stream`** for all file downloads via `getFile`. Since the header IS present (just wrong), the URL-extension fallback never runs.

The result: `ContentBlock::Image { media_type: "application/octet-stream", data: "..." }` is sent to the LLM, which rejects it because `application/octet-stream` is not a valid image MIME type.

## Evidence

Session logs show: `[image: application/octet-stream]` — confirming the image data arrives but with wrong MIME type.

## Proposed Fix

1. If `Content-Type` is `application/octet-stream` or another non-image type, fall through to secondary detection
2. Add magic-byte detection (JPEG `FF D8`, PNG `89 50 4E 47`, etc.) as the most reliable method
3. Keep URL extension as tertiary fallback

## Environment

- OpenFang v0.3.46
- Channel: Telegram (polling mode)
- LLM: qwen-plus (OpenAI-compatible API)

## Related

- #362 — Original issue (Telegram photos silently dropped)
- PR #365 — My original PR implementing incoming media support (closed without merge)

f-liva:
## Summary

- Telegram's file download API returns `Content-Type: application/octet-stream` for all files
- `download_image_to_blocks()` trusted this header, sending `ContentBlock::Image { media_type: "application/octet-stream" }` to LLM providers
- LLM providers (OpenAI, Claude, Qwen, etc.) reject `application/octet-stream` as invalid image MIME type, silently breaking vision/multimodal support
- Added magic-byte detection (JPEG `FF D8`, PNG `89 50 4E 47`, GIF `GIF8`, WEBP `RIFF...WEBP`) as the most reliable detection method
- Filter non-`image/*` Content-Type headers so URL-extension fallback can activate
- Extracted `media_type_from_url()` and `media_type_from_magic_bytes()` as reusable helpers

## Changes

- **bridge.rs**: `download_image_to_blocks()` — filter non-image Content-Type, add magic-byte detection after reading bytes
- **bridge.rs**: New `media_type_from_magic_bytes()` — detects JPEG, PNG, GIF, WebP from file signatures
- **bridge.rs**: New `media_type_from_url()` — extracted URL-extension fallback into standalone function
- **bridge.rs**: 8 new unit tests covering all magic-byte formats, unknown types, edge cases, and URL detection

## Test plan

- [x] `media_type_from_magic_bytes` correctly identifies JPEG, PNG, GIF, WebP
- [x] Returns `None` for unknown formats and too-short inputs
- [x] `media_type_from_url` correctly detects extensions in URLs
- [x] Telegram URLs without extensions default to `image/jpeg`
- [ ] Integration test: send photo via Telegram, verify `image/jpeg` media type reaches LLM

Fixes #528

### DIFF
--- a/crates/openfang-channels/src/bridge.rs
+++ b/crates/openfang-channels/src/bridge.rs
@@ -819,18 +819,11 @@ async fn download_image_to_blocks(url: &str, caption: Option<&str>) -> Vec<Conte
         .headers()
         .get("content-type")
         .and_then(|v| v.to_str().ok())
-        .map(|ct| ct.split(';').next().unwrap_or(ct).trim().to_string());
+        .map(|ct| ct.split(';').next().unwrap_or(ct).trim().to_string())
+        .filter(|ct| ct.starts_with("image/"));
 
     let media_type = content_type.unwrap_or_else(|| {
-        if url.contains(".png") {
-            "image/png".to_string()
-        } else if url.contains(".gif") {
-            "image/gif".to_string()
-        } else if url.contains(".webp") {
-            "image/webp".to_string()
-        } else {
-            "image/jpeg".to_string()
-        }
+        media_type_from_url(url)
     });
 
     let bytes = match resp.bytes().await {
@@ -855,6 +848,11 @@ async fn download_image_to_blocks(url: &str, caption: Option<&str>) -> Vec<Conte
         return vec![ContentBlock::Text { text: desc }];
     }
 
+    // Refine media type using magic bytes — most reliable detection method.
+    // Telegram (and some other services) return application/octet-stream for
+    // all file downloads, so the Content-Type header alone is not sufficient.
+    let media_type = media_type_from_magic_bytes(&bytes).unwrap_or(media_type);
+
     let data = base64::engine::general_purpose::STANDARD.encode(&bytes);
 
     let mut blocks = Vec::new();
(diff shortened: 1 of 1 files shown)


jaberjaber23:
Already on main with three-tier detection (Content-Type header → magic bytes → URL extension). Thanks.

PR #536, opened 2026-03-11
fix: suppress agent error posts on public broadcast channels (Mastodon)

Fail-Safe:
## Summary

- Adds `suppress_error_responses() -> bool` to the `ChannelAdapter` trait (default `false`)
- `MastodonAdapter` overrides to `true` — agent errors are logged internally but not posted as statuses
- Both `dispatch_text_message` and `dispatch_multimodal_message` error branches respect this flag

## Problem

When an agent hit a resource quota (e.g. token limit exceeded), the error was passed to `send_response()` which on Mastodon means calling `api_post_status()`. This caused hundreds of `"Agent error: Resource quota exceeded: Token limit exceeded"` posts to be published to a user's Mastodon account in a burst, all within the same second.

Errors are still logged via `warn!()` and recorded in delivery history — they're just no longer broadcast as public/unlisted statuses.

## Why this approach

Adding a trait method keeps the fix localized to adapters that opt in. All existing adapters (Telegram, Slack, Discord, etc.) are unaffected. Any future public broadcast adapters (e.g. Bluesky) can override `suppress_error_responses()` to `true` the same way.

## Test plan

- [ ] Verify `cargo build -p openfang-channels` passes cleanly
- [ ] Trigger an agent token-limit error with Mastodon channel enabled — confirm no status is posted
- [ ] Confirm the error is still present in delivery history / logs
- [ ] Verify Telegram/Slack/Discord still receive error responses as before

### DIFF
--- a/crates/openfang-channels/src/bridge.rs
+++ b/crates/openfang-channels/src/bridge.rs
@@ -770,14 +770,16 @@ async fn dispatch_message(
             send_lifecycle_reaction(adapter, &message.sender, msg_id, AgentPhase::Error).await;
             warn!("Agent error for {agent_id}: {e}");
             let err_msg = format!("Agent error: {e}");
-            send_response(
-                adapter,
-                &message.sender,
-                err_msg.clone(),
-                thread_id,
-                output_format,
-            )
-            .await;
+            if !adapter.suppress_error_responses() {
+                send_response(
+                    adapter,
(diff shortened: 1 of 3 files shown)


jaberjaber23:
Incorporated into main — suppress_error_responses() trait method added, Mastodon overrides to true.

PR #542, opened 2026-03-12
fix: Resolve channel default_agent by name at message-time instead of caching UUID at startup

psumo:
## Problem

When a channel's `default_agent` is resolved at startup, the UUID gets cached in the `AgentRouter`. Kill + respawn assigns a new UUID, but the router still holds the old one. Every message after that fails with `Agent not found: <old-uuid>` until you restart the whole process.

I kept hitting this while iterating on agent manifests:
1. Edit manifest (tweak system prompt, change model, etc)
2. `agent kill foo` + `agent spawn foo`
3. Agent comes back with a new UUID
4. Telegram messages all break — router is still pointing at the dead UUID
5. Have to restart the entire container to fix it

## Fix

Lazy re-resolution fallback in the bridge dispatch path. When `send_message()` gets an "Agent not found" error:

1. Look up the channel's configured default agent **name** (not UUID)
2. Call `find_agent_by_name()` to get the current UUID
3. Update the router cache
4. Retry the message

The cached UUID is still checked first so there's no overhead in the normal case. Re-resolution only kicks in when the send actually fails.

### `crates/openfang-channels/src/router.rs` (+24 lines)
- `channel_default_names: DashMap<String, String>` field — stores the configured agent name next to the cached UUID
- `set_channel_default_with_name()` — stores both name and ID
- `channel_default_name()` — getter for the stored name
- `update_channel_default()` — swap out just the cached UUID after re-resolution

### `crates/openfang-channels/src/bridge.rs` (+117 lines)
- `try_reresolution()` — checks for "Agent not found", looks up name from router, calls `find_agent_by_name()`, updates cache, returns new ID
- `dispatch_message()` error path — on "Agent not found", tries re-resolution + one retry
- `dispatch_with_blocks()` — same logic for multimodal messages
- `blocks` → `blocks.clone()` so blocks can be reused on retry

### `crates/openfang-api/src/channel_bridge.rs` (1 line)
- `set_channel_default()` → `set_channel_default_with_name()` at startup so the name is preserved

## Backwards compat

- `set_channel_default()` is untouched — only the startup path uses the new `_with_name` variant
- No new trait methods — uses existing `find_agent_by_name()`
- No config or API changes
- New fields on `AgentRouter` are private
- Existing tests pass as-is

## Testing

1. Configure a channel with `default_agent = "foo"`
2. Send a message — works fine
3. `agent kill foo` → `agent spawn foo`
4. Send another message
5. Before: `Agent error: Agent not found: ...`
6. After: message goes through, logs show re-resolution happened
7. Subsequent messages hit the updated cache with no re-resolution

Edge cases:
- Agent killed but not respawned → falls through to the original error
- Non-"Agent not found" errors (model API failures etc) → no re-resolution triggered
- Multiple channels with different defaults → each re-resolves independently
- Multimodal messages → retry works with cloned blocks

### DIFF
--- a/crates/openfang-api/src/channel_bridge.rs
+++ b/crates/openfang-api/src/channel_bridge.rs
@@ -1603,7 +1603,7 @@ pub async fn start_channel_bridge_with_config(
                     "{} default agent: {name} ({agent_id}) [channel: {channel_key}]",
                     adapter.name()
                 );
-                router.set_channel_default(channel_key, agent_id);
+                router.set_channel_default_with_name(channel_key, agent_id, name.clone());
                 // First configured default also becomes system-wide fallback
                 if !system_default_set {
                     router.set_default(agent_id);
--- a/crates/openfang-channels/src/bridge.rs
+++ b/crates/openfang-channels/src/bridge.rs
@@ -434,6 +434,65 @@ async fn send_lifecycle_reaction(
     let _ = adapter.send_reaction(user, message_id, &reaction).await;
 }
 
+/// Attempt to re-resolve an agent by channel default name after an "Agent not found" error.
+///
+/// When an agent is killed and respawned, it gets a new UUID. The router's cached ID
+/// becomes stale. This function:
+/// 1. Checks if the error indicates the agent was not found.
+/// 2. Looks up the channel's configured default agent name.
+/// 3. Calls `find_agent_by_name()` to get the new UUID.
+/// 4. Updates the router's channel default cache.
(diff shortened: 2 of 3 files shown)


jaberjaber23:
Good concept — re-resolution by name when cached UUID is stale. Has merge conflicts though. Please rebase onto main.

psumo:
Rebased onto current main and opened a clean PR: #626. This one should merge cleanly.

PR #547, opened 2026-03-12
feat(cli): externalize agent templates and seed defaults into ~/.open…

lvy010:
## Summary
- Remove binary fallback for agent templates; load templates from filesystem directories.
- Seed `~/.openfang/agents` from discovered seed dirs without overwriting user files.
- Seed `~/.openfang/catalog/*.json` on init when missing to preserve zero-break startup.

## Test Plan
- [√] Clean init creates `agents/` and `catalog/`
- [√] Remove all `~/.openfang/agents/*` then run `openfang agent new`
- [√] Verify output shows `No agent templates found` (no embedded fallback list)

### DIFF
--- a/crates/openfang-cli/src/bundled_agents.rs
+++ b/crates/openfang-cli/src/bundled_agents.rs
@@ -1,56 +1,71 @@
-//! Compile-time embedded agent templates.
+//! Filesystem-backed agent template seeding.
 //!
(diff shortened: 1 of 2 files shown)


jaberjaber23:
The compile-time embedded templates via include_str!() are intentional — they ensure openfang agent new works immediately after install with zero filesystem setup. Moving to filesystem discovery would break this. Users can already customize agent templates after creation.

PR #548, opened 2026-03-12
test(catalog): cover home-backed catalog bootstrapping behavior

lvy010:
## Summary
- Add focused tests for home-backed catalog behavior.
- Verify missing catalog files return an empty runtime catalog.
- Verify seeding populates catalog files and load succeeds.

## Test Plan
- [√] `cargo test -p openfang-runtime model_catalog:: -- --nocapture`


jaberjaber23:
from_home_dir() is defined and tested but never called from the kernel boot path or anywhere else. The existing catalog is built in-memory from compiled-in data + custom_models.json. This would be dead code. If you'd like to wire it into the kernel boot path as an alternative to the builtin catalog, please resubmit with the integration included.

PR #561, opened 2026-03-12
fix: complete Telegram image pipeline for Claude Code agents

### LINKED ISSUE #528: bug(channels): Telegram image media_type is application/octet-stream — breaks vision

f-liva:
## Summary

Fixes the end-to-end Telegram image pipeline so photos sent via Telegram reach Claude Code agents correctly and are processed with vision capabilities.

**Three bugs were found and fixed:**

1. **Vision model selection logic** (`kernel.rs`): When an image arrived, the kernel unconditionally swapped to a `vision_model` from config — even when the agent's current model (e.g. Claude Opus) already supports vision natively. This caused auth errors when the fallback vision model had misconfigured credentials. Now uses priority-based selection:
   - Priority 1: Explicit `vision_model` from config (forced override)
   - Priority 2: Current agent model if it supports vision (no swap)
   - Priority 3: Warning — no vision capability available

2. **Wrong CLI flag** (`claude_code.rs`): The driver used `--files` (plural) which doesn't exist in Claude Code CLI. Changed to `--file` (singular).

3. **`--file` requires Files API** (`claude_code.rs`): Even with the correct flag, `--file` expects Anthropic Files API file IDs and requires `CLAUDE_CODE_SESSION_ACCESS_TOKEN`. Replaced with `@path` syntax — Claude Code's native way to reference local files inline in the prompt. Images are now written to temp files and referenced as `@/tmp/openfang-img-xxx.jpg` in the prompt text.

## Changes

- `crates/openfang-kernel/src/kernel.rs` — Priority-based vision model selection with model catalog lookup
- `crates/openfang-runtime/src/drivers/claude_code.rs` — Rewrote image handling: temp file extraction, `@path` prompt injection, cleanup after CLI exits, refactored `build_args()` for testability
- Added unit tests for `build_args()` covering skip-permissions, streaming, and model flags

## Test plan

- [x] Send photo via Telegram to Claude Code agent (Opus) → agent sees and describes the image
- [x] No auth errors, no session token requirements
- [x] Temp files cleaned up after processing
- [x] Unit tests pass for build_args and build_prompt

### DIFF
--- a/crates/openfang-runtime/src/drivers/claude_code.rs
+++ b/crates/openfang-runtime/src/drivers/claude_code.rs
@@ -92,8 +92,8 @@ impl ClaudeCodeDriver {
 
     /// Build a text prompt from the completion request messages.
     ///
-    /// Image content blocks are represented as `[Attached image: <filename>]`
-    /// placeholders — the actual image files are passed via `--file`.
+    /// Image content blocks are referenced using Claude Code's `@path` syntax,
+    /// which tells the CLI to read the local file directly.
     fn build_prompt(request: &CompletionRequest, image_files: &[PathBuf]) -> String {
         let mut parts = Vec::new();
         let mut img_idx = 0;
@@ -127,11 +127,8 @@ impl ClaudeCodeDriver {
                             }
                             ContentBlock::Image { .. } | ContentBlock::ImageUrl { .. } => {
                                 if img_idx < image_files.len() {
-                                    let fname = image_files[img_idx]
-                                        .file_name()
-                                        .map(|n| n.to_string_lossy().to_string())
-                                        .unwrap_or_else(|| format!("image_{img_idx}"));
-                                    msg_parts.push(format!("[Attached image: {fname}]"));
+                                    let path = &image_files[img_idx];
+                                    msg_parts.push(format!("@{}", path.display()));
                                     img_idx += 1;
                                 }
                             }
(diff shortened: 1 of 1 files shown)


jaberjaber23:

PR #567, opened 2026-03-12
feat(telegram): Reply-to-message context for quoted messages

### LINKED ISSUE #564: feat: Reply-to-message support in Telegram bridge
## Summary

Add support for Telegram's `reply_to_message` field so the agent understands which message the user is quoting/replying to.

## Problem

When a user replies to a specific message in Telegram (using the quote/reply feature), the agent receives only the new message text without any context about which previous message is being referenced. This makes it impossible for the agent to understand conversational context from quoted replies.

### Telegram adapter (`crates/openfang-channels/src/telegram.rs`)
- Extract from `reply_to_message` JSON field:
  - `reply_to_text` — text of the quoted message
  - `reply_to_caption` — caption if the quoted message was a photo/file
  - `reply_to_sender` — who sent the original message
  - `reply_to_message_id` — ID of the quoted message

### Bridge (`crates/openfang-channels/src/bridge.rs`)
- Prepend quoted context before the user's message when dispatching to the agent

## Expected Behavior

When a user quotes a message in Telegram, the agent should receive:

```
[Replying to Federico: text of the quoted message]

The new message here...
```

Should work for both text messages and image/file captions.

f-liva:
## Summary
- Extracts `reply_to_message` metadata from Telegram updates (text, caption, sender, message_id)
- Prepends reply context as a contextual prefix when dispatching to agents
- Supports both text and image message paths

## Motivation
When users reply/quote a previous message in Telegram, the agent should see what message is being referenced to provide contextually relevant responses.

Closes #564

## Test plan
- [x] Reply to a message in Telegram — agent sees the quoted content
- [x] Reply to own message — context correctly shows sender name
- [x] Image with reply — reply context prepended before image blocks
- [x] Regular messages (no reply) — no change in behavior

### DIFF
--- a/crates/openfang-channels/src/bridge.rs
+++ b/crates/openfang-channels/src/bridge.rs
@@ -434,6 +434,29 @@ async fn send_lifecycle_reaction(
     let _ = adapter.send_reaction(user, message_id, &reaction).await;
 }
 
+/// Build a contextual prefix when the user is replying to a previous message.
+///
+/// Returns `Some("[Replying to <sender>: <text>]\n\n")` if reply metadata exists.
+fn build_reply_context(metadata: &std::collections::HashMap<String, serde_json::Value>) -> Option<String> {
+    // Need at least some quoted content
+    let quoted = metadata
+        .get("reply_to_text")
+        .and_then(|v| v.as_str())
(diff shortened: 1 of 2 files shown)


jaberjaber23:
Already on main with a more complete implementation including 5 test cases. Thanks for the original contribution — it's been expanded.

PR #574, opened 2026-03-13
Enhance channel_send tool with improved LLM examples

Hypn0sis:
## Testing

## Security

- [x] No new unsafe code


jaberjaber23:
Changing required fields from [channel, recipient] to [channel, message] is a behavioral change that could break existing usage where recipient is mandatory. The original descriptions also include channel-specific details (which channels support image_url, etc.) that would be lost. Closing for now — if you'd like to resubmit with just the examples added to the description without changing required fields, that would work.

PR #585, opened 2026-03-13
fix: prevent memory section misdirection when memories are already injected

### LINKED ISSUE #583: Memory section always directs model to call memory_recall, even when memories are already injected
## Bug Description

`build_memory_section()` in `crates/openfang-runtime/src/prompt_builder.rs` unconditionally emits the instruction:

> "When the user asks about something from a previous conversation, use memory_recall first."

This instruction is emitted regardless of whether recalled memories have already been injected into the system prompt via the `memories` parameter.

## Expected Behavior

When memories are already present in the prompt (i.e., the `memories` slice is non-empty), the model should be instructed to **use those memories directly** rather than calling `memory_recall` again.

## Actual Behavior

The model receives recalled memories in its context AND an instruction to "use memory_recall first." Models that follow instructions faithfully — which is most of them — obey the directive, call `memory_recall` against the KV store, receive nothing (the KV store doesn't contain the same data as the pre-injected semantic recall results), and report "no memory found" to the user.

The memories are right there in the context window, unused.

## Reproduction

1. Configure an agent with populated semantic memory (e.g., 100+ stored memories)
2. Send a message that triggers memory recall (e.g., "What do you remember about X?")
3. Observe that the system prompt contains a `## Memory` section with `Recalled memories:` entries
4. Observe that the model still calls `memory_recall` tool instead of using the injected memories
5. The `memory_recall` tool queries the KV store, which returns nothing useful
6. Model reports "I don't have any memories about that" despite the data being in its context

## Impact

This affects every model and every provider. Tested and confirmed on:
- **xAI Grok 4.1-fast** via OpenRouter
- **DeepSeek V3** via OpenRouter  
- **Qwen 3.5 Plus** via OpenRouter

The bug is particularly harmful for agents with rich memory stores (thousands of entries) because the semantic recall system works correctly — it finds and injects relevant memories — but the model ignores them due to the conflicting instruction.

## Root Cause

In `prompt_builder.rs`, the `build_memory_section()` function (around line 277):

```rust
pub fn build_memory_section(memories: &[(String, String)]) -> String {
    let mut out = String::from(
        "## Memory\n\
         - When the user asks about something from a previous conversation, use memory_recall first.\n\
         - Store important preferences, decisions, and context with memory_store for future use.",
    );
    if !memories.is_empty() {
        out.push_str("\n\nRecalled memories:\n");
        // ... appends memories ...
    }
    out
}
```

The "use memory_recall first" instruction is always emitted before the memories are appended. The model sees both the instruction and the memories, and follows the instruction.

## Proposed Fix

Make the instruction conditional on whether memories are present:

- **No memories** → existing behavior ("use memory_recall first")  
- **Memories present** → "The following memories were automatically recalled. Use them directly — do not call memory_recall for information already shown here."

A PR with this fix, including updated tests, is forthcoming.

## Environment

- OpenFang v0.3.46 through v0.4.0
- Bug has been present since at least v0.3.9 (earliest version tested)
- The fix has been running in production across 4 OpenFang instances for 48+ hours with no regressions

---

*Reported by Cairn-2001 (Cairn-2001@smoothcurves.nexus), OpenFang maintainer for the HACS coordination system at smoothcurves.nexus*

LupoGrigi0:
## Summary

Fixes #583 — `build_memory_section()` unconditionally tells the model to "use memory_recall first" even when recalled memories are already injected into the system prompt.

Models follow this instruction faithfully, call `memory_recall` against the KV store, get nothing back, and report "no memory found" — while the actual memories sit unused in their context window.

## Changes

**File:** `crates/openfang-runtime/src/prompt_builder.rs`

The fix makes the instruction conditional:

| Condition | Before (broken) | After (fixed) |
|-----------|-----------------|---------------|
| No memories injected | "use memory_recall first" | "use memory_recall first" *(unchanged)* |
| Memories present | "use memory_recall first" + memories appended below | "Use them directly — only call memory_recall for specific key lookups not covered below" |

### Code change

```rust
pub fn build_memory_section(memories: &[(String, String)]) -> String {
    let mut out = String::from("## Memory\n");
    if memories.is_empty() {
        // No recalled memories — direct model to use memory tools
        out.push_str(
            "- When the user asks about something from a previous conversation, use memory_recall first.\n\
             - Store important preferences, decisions, and context with memory_store for future use.",
        );
    } else {
        // Recalled memories are already injected — tell model to USE them directly
        out.push_str(
            "The following memories were automatically recalled based on the current conversation. \
             Use them directly in your responses — do not call memory_recall for information \
             already shown here. Only use memory_recall for specific key lookups not covered below.\n\n\
             Recalled memories:\n",
        );
        // ... existing memory formatting ...
    }
    out
}
```

## Testing

- All 32 existing `prompt_builder` tests pass
- Updated test assertions to verify the conditional behavior:
  - Empty memories → contains "use memory_recall first", does NOT contain "Use them directly"
  - With memories → contains "Use them directly", does NOT contain "use memory_recall first"
- Zero clippy warnings
- Live-tested across 4 production OpenFang instances (v0.3.46 → v0.4.0) for 48+ hours
- Confirmed fix on Grok 4.1-fast, DeepSeek V3, and Qwen 3.5 Plus via OpenRouter

## Impact

This bug affects every model and provider. With v0.4.0 adding support for 13 tool call recovery patterns and new drivers (Qwen Code, expanded Gemini), more models than ever can pilot OpenFang — and every one of them will hit this misdirection bug if they have populated memory stores.

The fix is minimal (one function, conditional branch) and backward-compatible. Agents with no memories see identical behavior.

---

*Submitted by Cairn-2001 (Cairn-2001@smoothcurves.nexus), OpenFang maintainer for the HACS (Human-Adjacent Coordination System) at smoothcurves.nexus*

### DIFF
--- a/crates/openfang-runtime/src/prompt_builder.rs
+++ b/crates/openfang-runtime/src/prompt_builder.rs
@@ -274,14 +274,29 @@ pub fn build_canonical_context_message(ctx: &PromptContext) -> Option<String> {
 /// Build the memory section (Section 4).
 ///
 /// Also used by `agent_loop.rs` to append recalled memories after DB lookup.
+///
+/// When recalled memories are present, the instruction changes: instead of
+/// directing the model to call `memory_recall` (which queries the KV store),
+/// it tells the model to use the already-injected memories directly. This fixes
+/// a misdirection bug where the model ignored injected memories and instead
+/// called `memory_recall`, got nothing from the empty KV store, and reported
+/// "no memory found".
 pub fn build_memory_section(memories: &[(String, String)]) -> String {
-    let mut out = String::from(
-        "## Memory\n\
-         - When the user asks about something from a previous conversation, use memory_recall first.\n\
-         - Store important preferences, decisions, and context with memory_store for future use.",
-    );
-    if !memories.is_empty() {
-        out.push_str("\n\nRecalled memories:\n");
+    let mut out = String::from("## Memory\n");
+    if memories.is_empty() {
+        // No recalled memories — direct model to use memory tools
+        out.push_str(
+            "- When the user asks about something from a previous conversation, use memory_recall first.\n\
+             - Store important preferences, decisions, and context with memory_store for future use.",
+        );
+    } else {
+        // Recalled memories are already injected — tell model to USE them directly
+        out.push_str(
+            "The following memories were automatically recalled based on the current conversation. \
+             Use them directly in your responses — do not call memory_recall for information \
+             already shown here. Only use memory_recall for specific key lookups not covered below.\n\n\
+             Recalled memories:\n",
+        );
         for (key, content) in memories.iter().take(5) {
             let capped = cap_str(content, 500);
             if key.is_empty() {
@@ -730,6 +745,7 @@ mod tests {
         assert!(section.contains("## Memory"));
         assert!(section.contains("memory_recall"));
         assert!(!section.contains("Recalled memories"));
+        assert!(!section.contains("Use them directly"));
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


jaberjaber23:
Already incorporated into main. Thanks for the fix.

PR #600, opened 2026-03-14
fix: agent provider not editable or correctly saved

Fail-Safe:
## Summary

Two bugs in the agent model/provider change flow:

- **Provider row was read-only** — no way to edit the provider independently from the model
- **Provider not saved for custom/local providers** (e.g. `mlx-lm-lg`, `mlx-lm-sm`) — entering `mlx-lm-lg/mlx-community/SomeModel` left the old provider unchanged and stored the full string as the model name

## Root Cause (provider not saved)

`infer_provider_from_model` checked the first-slash prefix against a hardcoded list of known providers. Custom/local provider names were not on the list, so provider resolved to `None` — skipping `update_model_and_provider` and storing the full unparsed string as the model.

HuggingFace repo IDs always have exactly one slash (`namespace/model`), so any input with **two or more slashes** unambiguously has an explicit provider prefix as its first segment. The fix returns it directly without requiring a match against the known-provider list.

## Changes

- `kernel.rs`: in `infer_provider_from_model`, treat the first segment as the provider when two or more slashes are present
- `index_body.html`: add inline Change editor to the Provider row (mirrors the existing Model row)
- `agents.js`: add `editingProvider`/`newProviderValue` state and `changeProvider()` method

## Test plan

- [ ] Click Change on the Provider row — inline editor appears pre-filled with current provider
- [ ] Change provider to `mlx-lm-lg`, save — provider updates correctly, toast confirms
- [ ] Click Change on the Model row and enter `mlx-lm-lg/mlx-community/SomeModel` — both provider and model update correctly (no more `Repo id must be in the form` error)
- [ ] Standard single-provider models (e.g. `groq/llama-3.3-70b-versatile`) still work

### DIFF
--- a/crates/openfang-api/static/index_body.html
+++ b/crates/openfang-api/static/index_body.html
@@ -922,7 +922,23 @@
                       </select>
                     </div>
                     <div class="detail-row" x-show="detailAgent.profile"><span class="detail-label">Profile</span><span class="detail-value" style="text-transform:capitalize" x-text="detailAgent.profile || '-'"></span></div>
-                    <div class="detail-row"><span class="detail-label">Provider</span><span class="detail-value" x-text="detailAgent.model_provider"></span></div>
+                    <div class="detail-row"><span class="detail-label">Provider</span>
+                      <template x-if="!editingProvider">
+                        <span>
+                          <span class="detail-value" x-text="detailAgent.model_provider"></span>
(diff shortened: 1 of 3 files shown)


jaberjaber23:
Incorporated into main — provider is now editable in the dashboard and multi-slash model names are handled correctly. Thanks.

PR #601, opened 2026-03-14
Fix cron channel delivery (was a no-op)

vnz:
## Summary

- **Fix no-op delivery:** `CronDelivery::Channel` only persisted metadata but never sent messages — cron jobs with `delivery: {kind: "channel"}` silently dropped all output
- **Delegate to channel adapters:** Deliver via the registered `send_channel_message()`, reusing existing `SlackAdapter`/`MatrixAdapter` with their connection pooling, auth, message splitting, and API version handling — works for any channel adapter registered at startup
- **Propagate delivery failures:** `record_success` now runs **after** delivery, so one-shot jobs are not removed on failure and recurring jobs show the error instead of false "ok" status

## Test plan

- [x] Unit test: `test_cron_delivery_metadata_persisted` — verifies `CronDelivery::Channel` variant fields
- [x] Clippy: zero warnings
- [x] CI: Check passed on all 3 platforms, Clippy passed, Tests passed on macOS/Windows (ubuntu runner was killed by infrastructure, not by a failing test; Format and Security Audit failures are pre-existing on main)
- [x] E2E: created cron job with Slack delivery → message appeared in channel → logs confirm delivery before success recording


jaberjaber23:
Good catch — CronDelivery::Channel was indeed a no-op. Incorporated your fix into main with two additional improvements: (1) LastChannel variant now actually delivers too (same bug, different path), (2) Webhook errors are now properly propagated instead of swallowed. Thanks for finding this.

vnz:
Thanks @jaberjaber23 — glad the fix was useful. Good call extending it to LastChannel and Webhook, those had the same swallowed-error pattern. Happy to contribute more.

PR #604, opened 2026-03-14
fix(whatsapp-gateway): reply to group messages in group, not as private DM

f-liva:
## Summary

- **Group reply routing**: When a message arrives from a WhatsApp group (`@g.us`), the gateway now replies in the same group instead of sending a private DM to the individual sender
- **Correct sender extraction**: For group messages, the real sender is extracted from `msg.key.participant` instead of `remoteJid` (which is the group JID)
- **Dynamic agent resolution**: Agent name is resolved to UUID via `/api/agents` endpoint instead of being hardcoded
- **Group metadata forwarding**: Group context (`group_jid`, `group_name`, `is_group`) is included in the metadata sent to OpenFang
- **JID-aware `sendMessage`**: The outbound `sendMessage` function now accepts full JIDs (including `@g.us` groups) in addition to phone numbers
- **Auto-connect on startup**: Gateway automatically reconnects if credentials exist from a previous session

## Test plan

- [x] Tested live with WhatsApp group "Liva's" — messages received and replies sent correctly in group
- [x] Verified DM (private) messages still work as before
- [x] Verified sender identity metadata is correctly extracted for both group and DM contexts

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/packages/whatsapp-gateway/index.js
+++ b/packages/whatsapp-gateway/index.js
@@ -9,7 +9,8 @@ const { randomUUID } = require('node:crypto');
 // ---------------------------------------------------------------------------
 const PORT = parseInt(process.env.WHATSAPP_GATEWAY_PORT || '3009', 10);
 const OPENFANG_URL = (process.env.OPENFANG_URL || 'http://127.0.0.1:4200').replace(/\/+$/, '');
-const DEFAULT_AGENT = process.env.OPENFANG_DEFAULT_AGENT || 'assistant';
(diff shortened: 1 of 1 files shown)


jaberjaber23:
Incorporated the group reply fix into main. Kept DEFAULT_AGENT as 'assistant' instead of 'ambrogio'. Also skipped the agent name→UUID resolver since the existing flow already handles name-based routing. Good catch on the group reply routing.

PR #606, opened 2026-03-14
feat(whatsapp-gateway): add media message support

### LINKED ISSUE #605: WhatsApp gateway drops media-only messages (images, voice, documents)
## Problem

The WhatsApp gateway silently drops all messages that don't contain text. When a user sends an image without a caption, a voice note, a document, or a sticker, the gateway skips it with `if (!text) continue;` — no error, no log.

## Root Cause

In `packages/whatsapp-gateway/index.js`, the message handler only extracts text from `conversation`, `extendedTextMessage.text`, and `imageMessage.caption`. If all are empty, the message is discarded.

## Expected Behavior

Media messages should be downloaded via Baileys' `downloadMediaMessage()`, uploaded to OpenFang's `/api/agents/{id}/upload` endpoint, and forwarded with the `attachments` array in the message payload so the LLM can see the content.

## Supported Media Types

- Images (with/without caption)
- Videos (with/without caption)  
- Audio/voice notes
- Documents (PDF, etc.)
- Stickers

f-liva:
## Summary

Closes #605

- Detect media messages (images, video, audio, documents, stickers) in incoming WhatsApp messages
- Download media via Baileys' `downloadMediaMessage()` 
- Upload to OpenFang's existing `/api/agents/{id}/upload` endpoint
- Forward the `file_id` in the `attachments` array so the LLM receives the content
- Graceful fallback: if media download/upload fails, text/caption is still forwarded

## Test plan

- [ ] Send image without caption → agent receives and describes the image
- [ ] Send image with caption → agent receives both image and caption text
- [ ] Send voice note → agent receives audio attachment
- [ ] Send document (PDF) → agent receives document attachment
- [ ] Send text-only message → behavior unchanged

### DIFF
--- a/packages/whatsapp-gateway/index.js
+++ b/packages/whatsapp-gateway/index.js
@@ -5,7 +5,7 @@ import { randomUUID } from 'node:crypto';
 import fs from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
-import makeWASocket, { useMultiFileAuthState, DisconnectReason, fetchLatestBaileysVersion } from '@whiskeysockets/baileys';
+import makeWASocket, { useMultiFileAuthState, DisconnectReason, fetchLatestBaileysVersion, downloadMediaMessage } from '@whiskeysockets/baileys';
 import QRCode from 'qrcode';
 import pino from 'pino';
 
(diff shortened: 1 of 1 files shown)


jaberjaber23:

PR #607, opened 2026-03-14
Add Parakeet MLX local speech support

yaroslavyaroslav:
## Summary

Adds local Parakeet MLX speech support to media understanding and registers the bundled hand configuration.

## Changes

- Add Parakeet handling in media understanding runtime flow.
- Register the bundled clip hand updates needed for Parakeet support.
- Wire the hand registry so the new local speech path is available.

## Testing

## Security

- [x] No new unsafe code


jaberjaber23:
Incorporated Parakeet MLX support into main. Changed auto-detection to require explicit OPENFANG_ENABLE_PARAKEET_MLX env var (cloud providers like Groq should take priority by default, not be overridden just because uv is installed). Thanks.

PR #609, opened 2026-03-14
Handle existing Chrome installs in browser setup

yaroslavyaroslav:
## Summary

Improves browser setup handling so existing Chrome installs are detected and managed correctly.

## Changes

- Update API route handling for browser setup around existing Chrome installs.
- Improve setup behavior when Chrome is already present on the machine.
- Preserve the existing browser-hand flow while tightening install detection.

## Testing

## Security

- [x] No new unsafe code

Fixes: #611


jaberjaber23:
The Chrome install handling fix is already on main (already an App at pattern included). Thanks.

PR #615, opened 2026-03-15
Fix silent/NO_REPLY responses leaking debug fallback to channels

### LINKED ISSUE #614: Silent/NO_REPLY responses leak debug fallback message to channels
## Bug

When an agent intentionally returns `NO_REPLY` (empty response with `silent: true`), the system generates a debug fallback message like:

```
[The agent completed processing but returned no text response. (3 in / 8 out | 1 iter)]
```

This message is sent back to the channel (WhatsApp, Telegram, etc.) instead of being suppressed.

## Root Cause

- `routes.rs`: The REST endpoint checks for empty responses but doesn't check the `silent` field from `AgentLoopResult`, generating a debug fallback instead.
- `ws.rs`: The WebSocket handler has the same issue — empty responses get a fallback message even when the agent chose to be silent.
- `types.rs`: `MessageResponse` struct was missing the `silent` field.

## Expected Behavior

When an agent returns `NO_REPLY` / silent response:
1. The API should return `{ "silent": true }` in the response
2. Channel gateways should not send any message to the user

## Affected Channels

- WhatsApp (gateway sends the debug text as a regular message)
- Any channel using the REST or WebSocket API

f-liva:
## Summary

Fixes #614 — Silent agent responses (NO_REPLY) were leaking a debug fallback message to WhatsApp and other channels instead of being suppressed.

### Changes

- **`routes.rs`**: Check `result.silent` before generating fallback; return `silent: true` in JSON response
- **`ws.rs`**: Respect empty responses as intentionally silent instead of injecting debug text
- **`types.rs`**: Add `silent` field to `MessageResponse` struct
- **`whatsapp-gateway/index.js`**: Skip sending message when API response indicates silent

## Test Plan

- [ ] Send a message to an agent configured with NO_REPLY behavior
- [ ] Verify no debug fallback message appears on WhatsApp
- [ ] Verify normal agent responses still work correctly
- [ ] Verify WebSocket clients receive the `silent` field

Closes #614

### DIFF
--- a/crates/openfang-api/src/routes.rs
+++ b/crates/openfang-api/src/routes.rs
@@ -403,17 +403,24 @@ pub async fn send_message(
             // Strip <think>...</think> blocks from model output
             let cleaned = crate::ws::strip_think_tags(&result.response);
 
-            // Guard: ensure we never return an empty response to the client
-            let response = if cleaned.trim().is_empty() {
-                format!(
-                    "[The agent completed processing but returned no text response. ({} in / {} out | {} iter)]",
-                    result.total_usage.input_tokens,
-                    result.total_usage.output_tokens,
-                    result.iterations,
-                )
-            } else {
-                cleaned
-            };
+            // If the agent intentionally chose not to reply (NO_REPLY / [[silent]]),
+            // OR if the response is empty after stripping <think> tags (agent only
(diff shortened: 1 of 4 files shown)


jaberjaber23:

PR #248, opened 2026-03-03
feat(runtime): add openai-responses provider and Responses API routing

mxyhi:
## Summary
- add a new provider alias: `openai-responses`
- route OpenAI driver to `/v1/responses` when model format requires Responses API
- allow explicit force format via `openai-responses/<model>` prefix
- support `OPENAI_REASONING_EFFORT` (`low|medium|high|xhigh`) and model suffix fallback
- update docs for provider/configuration usage

## Implementation
- `drivers/mod.rs`: register `openai-responses` and include it in known provider hints
- `drivers/openai.rs`: add Responses request/response structs and routing logic
- `agent_loop.rs`: prepend `openai-responses/` to `api_model` when provider is `openai-responses`

## Tests
- `cargo check -p openfang-runtime`
- `cargo test -p openfang-runtime drivers::openai::tests:: -- --nocapture`
- `cargo test -p openfang-runtime test_known_providers_list -- --nocapture`

PR #628, opened 2026-03-15
feat(runtime): add openai-responses provider and Responses API routing

mxyhi:
Supersedes #248.

GitHub refused to reopen #248 (`Could not open the pull request`), while `origin/main` still does not contain this feature and this branch has been rebased onto current `main`.

## Summary
- add a new provider alias: `openai-responses`
- route the OpenAI driver to `/v1/responses` when model format requires the Responses API
- allow explicit force format via `openai-responses/<model>` prefix
- support `OPENAI_REASONING_EFFORT` and model suffix fallback
- update provider/configuration docs

## Implementation
- `crates/openfang-runtime/src/drivers/mod.rs`: register `openai-responses` and expose it in known provider hints
- `crates/openfang-runtime/src/drivers/openai.rs`: add Responses request/response structs and routing logic on top of current mainline driver behavior
- `crates/openfang-runtime/src/agent_loop.rs`: prepend `openai-responses/` to `api_model` when provider is `openai-responses`
- `docs/configuration.md` / `docs/providers.md`: document provider selection and `OPENAI_REASONING_EFFORT`

## Verification
- `cargo test -p openfang-runtime openai::tests -- --nocapture`
- `cargo test -p openfang-runtime test_known_providers_list -- --nocapture`

### DIFF
--- a/crates/openfang-runtime/src/agent_loop.rs
+++ b/crates/openfang-runtime/src/agent_loop.rs
@@ -338,6 +338,11 @@ pub async fn run_agent_loop(
 
         // Strip provider prefix: "openrouter/google/gemini-2.5-flash" → "google/gemini-2.5-flash"
         let api_model = strip_provider_prefix(&manifest.model.model, &manifest.model.provider);
+        let api_model = if manifest.model.provider == "openai-responses" {
+            format!("openai-responses/{api_model}")
+        } else {
+            api_model
+        };
 
         let request = CompletionRequest {
             model: api_model,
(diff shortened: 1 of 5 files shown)


jaberjaber23:
Reviewed. The ResponsesRequest struct uses hallucinated fields (instructions, input) that don't match OpenAI's actual Responses API schema. The build_responses_input() converts messages to plain text instead of the structured JSON format the real API expects. Model detection targets gpt-5.3-codex which doesn't exist yet. This would fail at runtime with 400 errors against the real OpenAI endpoint. Please verify the request/response format against https://platform.openai.com/docs/guides/migrate-to-responses and resubmit with the correct schema.

PR #689, opened 2026-03-17
fix(runtime): retry retryable LLM errors + action validator

### LINKED ISSUE #688: Agent loop does not verify tool execution when user requests explicit actions
## Bug Description

When a user sends a message containing an explicit action request (e.g., "execute X and send to Telegram"), the LLM sometimes generates only a text summary without actually calling the required tool. The agent loop does not detect this discrepancy, treating the response as complete.

This leads to **"phantom actions"** — the agent replies "Done, sent!" but nothing was actually sent or executed.

## Root Cause

In `crates/openfang-runtime/src/agent_loop.rs`, the loop processes the LLM response and checks for tool calls. However, when the LLM responds with text only (`StopReason::EndTurn`), the system treats the turn as successfully complete. There is no validation step that cross-references the user's intent (containing action verbs like "send", "execute", "create", "delete" paired with a target like Telegram, WhatsApp, email) against the actual tool calls made during the turn.

## Key Files

- `crates/openfang-runtime/src/agent_loop.rs` — the main loop processing tool calls vs text responses
- `crates/openfang-runtime/src/tool_runner.rs` — tool execution logic
- `crates/openfang-runtime/src/channel_bridge.rs` — response routing to channels

## Steps to Reproduce

1. Send a message like "invia questo riepilogo su Telegram" (send this summary to Telegram)
2. The LLM generates a nicely formatted summary text and says "Done, sent!"
3. No `send_telegram_message` tool call appears in the logs
4. The message was never actually sent to Telegram

## Proposed Fix

Add an **action validator** step in the agent loop that:

1. Detects when the user's message contains explicit action keywords (`send`, `execute`, `create`, `delete`, `write`) paired with a target channel or service (Telegram, WhatsApp, email, calendar, etc.)
2. After the LLM responds, checks whether any matching tool calls were made during the turn
3. If the response contains no matching tool calls but the user clearly requested an action, re-prompts the model with a system reminder to actually execute the action rather than just describing it

This could be implemented as a post-response validation hook in `agent_loop.rs` before finalizing the turn.

## Related

See #677 for broader WhatsApp reliability context.

f-liva:
## Summary

Fixes #688 — Agent loop does not verify tool execution when user requests explicit actions.

- **Retry retryable LLM errors**: The catch-all `Err` branch in `call_with_retry` and `stream_with_retry` returned immediately on any error, ignoring the `is_retryable` flag from the error classifier. Timeouts and transient network errors (classified as retryable) were treated as fatal. Now they are retried with exponential backoff (using `suggested_delay_ms` from the provider when available).
- **Action validator**: When the user explicitly requests a side-effecting action (e.g. "send to Telegram", "execute the script and send") but the LLM responds with text only and no tool call, the loop re-prompts once to force tool execution. This prevents "phantom actions" where the agent says "Done, sent!" without actually calling any tool.

## Changes

**`crates/openfang-runtime/src/agent_loop.rs`**:
- Added `action_validator_retried` flag to both `run_agent_loop` and `run_agent_loop_streaming`
- Added action validator block that detects text-only responses when user requested explicit actions, and re-prompts the LLM
- Fixed `call_with_retry` and `stream_with_retry` to actually retry on retryable errors with exponential backoff
- Added `requires_tool_action()` function — conservative pattern matching for explicit action requests (supports Italian and English, matches action verb + target channel patterns)

## Test plan

- [ ] Verify `cargo build` passes (confirmed locally)
- [ ] Verify `cargo clippy` passes (confirmed locally)
- [ ] Test: send a message like "esegui lo script e invia su Telegram" — should execute tool, not just describe it
- [ ] Test: send a normal question — should NOT trigger the action validator
- [ ] Test: simulate an LLM timeout — should retry instead of returning "Request timed out"

Closes #688


f-liva:
Closing — no longer needed.

PR #692, opened 2026-03-17
fix: use dynamic providers in agent spawn form

lc-soft:
## Summary

This changes the provider field in the agent spawn form from a hardcoded list to a dynamically loaded list from /api/providers, so the UI always reflects the currently available providers.

## Changes

Before:

screenshot-20260317-175657

After:

screenshot-20260317-173515

## Testing

## Security

- [ ] No new unsafe code


lc-soft:
#817

PR #694, opened 2026-03-17
Feat/programmatic tool calling

skymoore:
## Summary

Adds programmatic tool calling (PTC) via a Python subprocess, saving 30-40% context usage by replacing verbose tool JSON schemas with concise Python function signatures in the agent's context:

```
web_fetch(url: str)
```

Instead of sending the full JSON schema for every tool, the agent receives a lightweight Python SDK and executes tool calls as Python function invocations through an IPC server.

## Future Direction: Monty

The current implementation shells out to a full CPython interpreter. As [Monty](https://github.com/pydantic/monty) matures — a standalone, statically-typed Python compiler built on Cranelift — it would be a natural replacement. Monty would eliminate the Python runtime dependency, enable embedding the interpreter directly in the OpenFang binary, and provide better sandboxing guarantees. Once Monty reaches sufficient stdlib and type-inference coverage, migrating PTC to use it instead of CPython should be straightforward since the generated SDK is already type-annotated.

## Changes

- Adds PTC system: SDK generator, IPC server, Python executor, tool classification
- PTC is enabled by default (can be disabled per-agent with `ptc_enabled: false`)
- Adds `execute_code` meta-tool that the LLM calls to run Python
- IPC server handles tool dispatch from Python subprocess back to the agent loop
- Includes fixes: heartbeat touch during PTC dispatch, phantom action detection, image block stripping, `has_any_content()` for thinking-only responses, `max_tick_duration_secs` config, extended thinking support
- Dockerfile updated to include Python 3.14

## Testing

- [x] `cargo fmt --check` passes
- [x] `cargo test --workspace` passes (2220+ tests, 0 failures)
- [x] Rebased cleanly on upstream/main


skymoore:
@jaberjaber23 ubuntu test was just randomly cancelled, so it needs to be re-run.  security audit issue is pre-existing.

jaberjaber23:
Reviewed. Real feature with sound architecture. Three changes needed: (1) Remove the 2,990-line auto-generated macOS-schema.json from the diff. (2) Python subprocess has no sandboxing, LLM-generated code runs with full daemon privileges. At minimum document this risk. (3) PTC should be opt-in (default false), not opt-out, for a feature that shells out to Python.

skymoore:
@jaberjaber23 change requests addressed.

Jengro777:

jaberjaber23:

PR #699, opened 2026-03-17
fix: preserve model/provider/temperature across Hand respawn

Fail-Safe:
## Problem

Every time a Hand agent was respawned (daemon restart, crash recovery, or manual reactivation), `activate_hand()` rebuilt the `AgentManifest` entirely from the compile-time-embedded `HAND.toml`, silently discarding any changes made via the agents API. This caused a persistent revert loop:

1. Patch `researcher-hand` via API (model, provider, tool filters) ✅
2. Daemon restart → `activate_hand()` rebuilds from `HAND.toml` → changes gone ❌
3. Patch again ✅
4. Crash recovery → same wipe ❌

### Commit 1 — preserve API-patched tool filters across Hand respawn

`activate_hand()` now reads the existing agent's `tool_allowlist` and `tool_blocklist` from the registry before rebuilding the manifest, and reapplies them after. Only applied when non-empty — an empty list means "no override set", not "block all".

### Commit 2 — preserve model/provider/temperature + temperature API

Extends the same pattern to model config fields. A single registry scan captures the existing agent's `provider`, `model`, and `temperature` before rebuild. After computing `hand_provider`/`hand_model` from `HAND.toml`, any difference from the live values indicates an API patch — those values are reapplied after the manifest build.

`system_prompt` is intentionally excluded: it is assembled dynamically from `HAND.toml` + settings context and must stay live.

Also adds:
- `registry::update_temperature()` for hot-patching sampling temperature
- `temperature` exposed in `GET /api/agents` and `GET /api/agents/{id}` responses
- `temperature` field in `PATCH /api/agents/{id}/config` with `0.0–2.0` validation
- Temperature input in the dashboard agent config tab

## What survives respawn after this PR

| Field | Before | After |
|---|---|---|
| `tool_allowlist` | ❌ | ✅ |
| `tool_blocklist` | ❌ | ✅ |
| `provider` | ❌ | ✅ |
| `model` | ❌ | ✅ |
| `temperature` | ❌ | ✅ |
| `system_prompt` | ❌ | ❌ intentional |

## Test plan

- [ ] Patch a Hand's model/provider via `PUT /api/agents/{id}/model`, restart daemon — agent should come back with the patched model
- [ ] Patch tool filters via `PUT /api/agents/{id}/tools`, restart — filters should be preserved
- [ ] Patch temperature via `PATCH /api/agents/{id}/config` with `{"temperature": 0.5}`, restart — temperature should be preserved
- [ ] Verify a Hand that was never patched still inherits `default_model` from config correctly (override path should be `None`)


jaberjaber23:
Reviewed and approved. Superset of #684 (which is now closed). Has merge conflicts. Please rebase onto current main.

jaberjaber23:

PR #711, opened 2026-03-18
fix(matrix): make auto_accept_invites configurable, default to false

Reaster0:
## Summary

- `MatrixAdapter` hardcoded `auto_accept_invites: true`, meaning any Matrix-connected instance would blindly join every room it was invited to
- This is a security concern for public-facing homeservers — a malicious user could invite the bot into an arbitrary room and interact with the agent without the operator's consent
- The field was not exposed in `MatrixConfig`, so operators had no way to disable it

## Changes

- Add `auto_accept_invites: bool` to `MatrixConfig` (`openfang-types`), with `#[serde(default)]` defaulting to `false`
- Thread the field through `MatrixAdapter::new()` instead of hardcoding `true`
- Wire it in `channel_bridge.rs` from `mx_config.auto_accept_invites`
- Update unit tests to pass the new parameter

### Migration

**Breaking change in default behavior**: existing setups that relied on auto-accept will need to explicitly opt in:

```toml
[channels.matrix]
auto_accept_invites = true
```

## Test plan

- [ ] `cargo build --workspace --lib` compiles
- [ ] `cargo test --workspace` passes (3 updated test call sites in `matrix.rs`)
- [ ] Start daemon with `auto_accept_invites = false` (default) — invite the bot to a room, verify it does **not** join
- [ ] Set `auto_accept_invites = true` — invite again, verify auto-join works
- [ ] Verify `allowed_rooms` filtering still applies when auto-accept is enabled

### DIFF
--- a/crates/openfang-api/src/channel_bridge.rs
+++ b/crates/openfang-api/src/channel_bridge.rs
@@ -1218,6 +1218,7 @@ pub async fn start_channel_bridge_with_config(
                 mx_config.user_id.clone(),
                 token,
                 mx_config.allowed_rooms.clone(),
+                mx_config.auto_accept_invites,
             ));
             adapters.push((adapter, mx_config.default_agent.clone()));
         }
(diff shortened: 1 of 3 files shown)


jaberjaber23 (approved):
Clean, minimal, security-positive change. Threads auto_accept_invites from config through channel_bridge to MatrixAdapter correctly. Default false is the right secure-by-default posture. Call chain verified complete. Tests updated.

PR #716, opened 2026-03-18
Add -d/--detach flag for background daemon mode

ReSyncc:
## Problem

OpenFang is an autonomous agent OS — it should run as a background service without requiring a dedicated terminal session. Currently, `openfang start` blocks the terminal and the daemon dies when the terminal is closed. Users have to manually background it with `nohup` or `&`, which is a poor experience for an always-on system.

This contradicts the core design: agents, channels (Telegram, Discord, etc.), scheduled jobs, and workflows should keep running independently of any terminal session.

## Solution

- **`openfang start`** now spawns the daemon in the background using the existing `start_daemon_background()` function (which was already used by `openfang dashboard` but not exposed to users) and returns immediately, printing the API URL and dashboard link
- **`openfang start`** when a daemon is already running now reports the existing daemon's URL instead of erroring out
- **`openfang start --foreground`** preserves the old blocking behavior for Docker containers and debugging
- **Dockerfile** updated to `CMD ["start", "--foreground"]` since containers need the foreground process to stay alive

## Benefits

- Users can run `openfang start` and close their terminal — the daemon keeps running
- Agents stay active, channels remain connected, scheduled jobs continue executing
- No more confusing "already running" errors — idempotent start behavior
- Docker still works correctly with explicit `--foreground`
- Aligns CLI behavior with what users expect from a daemon/service command

## Test plan
- [x] `cargo build -p openfang-cli` compiles
- [x] `openfang start` returns immediately, daemon accessible at printed URL
- [x] `openfang start` again prints "already running" with URL
- [ ] `openfang start --foreground` blocks as before
- [ ] `openfang stop` stops the background daemon
- [ ] Docker container stays alive with `start --foreground`

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ The entire system compiles to a **single ~32MB binary**. One install, one comman
 ```bash
 curl -fsSL https://openfang.sh/install | sh
 openfang init
-openfang start
+openfang start              # or `openfang start -d` to run in the background
 # Dashboard live at http://localhost:4200
 ```
 
@@ -54,7 +54,7 @@ openfang start
 ```powershell
 irm https://openfang.sh/install.ps1 | iex
 openfang init
-openfang start
+openfang start              # or `openfang start -d` to run in the background
 ```
 
 </details>
@@ -319,7 +319,7 @@ The gateway listens on port `3009` by default. Override with `WHATSAPP_GATEWAY_P
 **5. Start OpenFang:**
 
 ```bash
-openfang start
+openfang start              # or `openfang start -d` to run in the background
 # Dashboard at http://localhost:4200
 ```
 
@@ -413,7 +413,7 @@ curl -fsSL https://openfang.sh/install | sh
 # 2. Initialize — walks you through provider setup
 openfang init
 
-# 3. Start the daemon
+# 3. Start the daemon (add -d to run in the background)
 openfang start
 
 # 4. Dashboard is live at http://localhost:4200
@@ -435,7 +435,7 @@ openfang agent spawn coder
 ```powershell
 irm https://openfang.sh/install.ps1 | iex
 openfang init
-openfang start
+openfang start              # or `openfang start -d` to run in the background
 ```
 
 </details>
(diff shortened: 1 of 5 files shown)


jaberjaber23 (changes requested):
Breaking behavior change that needs more work:

1. `--config` / `-c` is not forwarded to the background process. If a user runs `openfang start -c /custom/config.toml`, the background process starts with default config, not the custom one.

2. Exit code contract change — "already running" changed from exit(1) to success return. Scripts checking exit codes will break silently.

3. This is a breaking change for every user, tutorial, and script that uses `openfang start` expecting foreground behavior. Needs discussion and a deprecation warning at minimum.

4. Test plan items are unchecked — the PR was not actually tested.

ReSyncc:
> Breaking behavior change that needs more work:
> 
>     1. `--config` / `-c` is not forwarded to the background process. If a user runs `openfang start -c /custom/config.toml`, the background process starts with default config, not the custom one.
> 
>     2. Exit code contract change — "already running" changed from exit(1) to success return. Scripts checking exit codes will break silently.
> 
>     3. This is a breaking change for every user, tutorial, and script that uses `openfang start` expecting foreground behavior. Needs discussion and a deprecation warning at minimum.
> 
>     4. Test plan items are unchecked — the PR was not actually tested.

Great point, ended up going to a `-d` command style similar to detach apart of docker run and docker compose to avoid breaking behavior.

- openfang start remains foreground (no breaking change)
- `--config` and `--yolo` flags forwarded to the background process
- "Already running" keeps exit(1) (preserves script behavior)
- Fixed all callers of start_daemon_background() (main.rs, init_wizard.rs, cmd_dashboard)
- Updated docs/cli-reference.md with new flag, examples, and output for both modes
- Added comment to `openfang start` in README.md
- Added tip apart of openfang start command to let users know there is a `-d` arg to detach.

ReSyncc:
- Daemon stdout/stderr redirected to ~/.openfang/daemon.log (append mode) instead of /dev/null
  - Unix: setsid() via pre_exec hook so daemon survives terminal close
  - libc added as Unix-only dependency for setsid
  - Detach success output now shows log file path
  - Timeout error message points user to daemon.log for debugging
  - Falls back to /dev/null if log file can't be opened

jaberjaber23:
Reviewed. Good feature but needs: (1) check setsid() return value, (2) add PID file to prevent orphan processes and double-start, (3) address TOCTOU race on double-start detection.

Jengro777:

jaberjaber23:

PR #732, opened 2026-03-18
fix: cross-channel message routing isolation

### LINKED ISSUE #731: Agent responds in wrong channel — cross-channel context contamination
## Bug description

An agent handling multiple channels (e.g. WhatsApp groups + private Telegram DMs) can confuse conversation contexts, responding in a WhatsApp group with a message clearly meant for a private chat with a different user.

**Observed behavior**: Agent "Ambrogio" was having a private technical conversation with user Federico about image debounce. Simultaneously, in a WhatsApp group, a different user (Anna) sent "Well, the pictures are interesting 😋". The agent interpreted Anna's group message as part of Federico's private conversation (keyword match on "pictures") and responded IN THE GROUP with a long technical message meant for the private chat.

### 1. Missing `ChannelContext` on image retry path

In `bridge.rs`, `dispatch_with_blocks()` calls `send_message_with_blocks()` **without** `ChannelContext` when retrying after agent re-resolution. The text path (`dispatch_message()`) correctly passes `ChannelContext` on retry — the image path doesn't.

Without context, the kernel builds the system prompt with no channel awareness section and no sender identity verification.

**File**: `crates/openfang-channels/src/bridge.rs` (dispatch_with_blocks retry path)

### 2. No per-channel message tagging in shared session

The canonical session is deliberately shared across all channels (cross-channel memory), but `Message` only has `role` + `content` — no channel/sender metadata. When the LLM sees session history, messages from a WhatsApp group and a private Telegram chat are indistinguishable, causing the LLM to confuse unrelated conversations that share keywords.

**File**: `crates/openfang-kernel/src/kernel.rs` (message injection into session)

## Expected behavior

1. Each message in the session should carry origin metadata so the LLM can distinguish conversations from different channels/users
2. `ChannelContext` must always be passed through all code paths (including retry/re-resolution)
3. Responses must never leak to a channel other than the one that triggered the message

## Related

- Separate from debounce issue (#728)

f-liva:
## Summary

- **Fix missing `ChannelContext` on image retry path** in `dispatch_with_blocks()`: the re-resolution path called `send_message_with_blocks()` without channel context, losing channel_type/sender_id/sender_name. Now uses `send_message_with_blocks_and_context()`, matching the text path.
- **Add per-channel message origin tagging**: user messages entering the session are now prefixed with `[via {channel}, from {sender}]` so the LLM can distinguish conversations from different channels/users in the shared canonical session. Applied to both streaming and non-streaming kernel paths.

Closes #731

## Test plan

- [ ] Verify `cargo build --workspace --lib` compiles
- [ ] Verify `cargo test -p openfang-channels` passes (9/9)
- [ ] Verify `cargo test -p openfang-runtime -- prompt_builder` passes (38/38)
- [ ] Manual test: send messages from two different channels to the same agent, verify responses don't cross channels
- [ ] Manual test: send an image that triggers agent re-resolution, verify response goes to the correct channel
- [ ] Verify session history shows `[via whatsapp, from X]` / `[via telegram, from Y]` tags on user messages

### DIFF
--- a/crates/openfang-channels/src/bridge.rs
+++ b/crates/openfang-channels/src/bridge.rs
@@ -1841,7 +1841,14 @@ async fn dispatch_with_blocks(
             // Try re-resolution before reporting error
             if let Some(new_id) = try_reresolution(&e, &channel_key, handle, router).await {
                 let typing_task2 = spawn_typing_loop(adapter_arc.clone(), message.sender.clone());
-                let retry = handle.send_message_with_blocks(new_id, blocks).await;
+                let ch_ctx2 = ChannelContext {
+                    channel_type: Some(ct_str.to_string()),
+                    sender_id: Some(sender_user_id(message).to_string()),
+                    sender_name: Some(message.sender.display_name.clone()),
+                };
+                let retry = handle
+                    .send_message_with_blocks_and_context(new_id, blocks, ch_ctx2)
+                    .await;
                 typing_task2.abort();
                 match retry {
                     Ok(response) => {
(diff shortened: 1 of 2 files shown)


f-liva:
Closing — no longer needed.

PR #749, opened 2026-03-19
fix: Add dynamic Git detection; fix Windows 'command not found' and e…

felix307253927:
…ncoding issues.

## Changes

feat: add dynamic Git Bash path detection
fix: resolve Windows encoding/garbled text issues
fix: handle "command not found" errors on Windows
fix: ensure Git built-in commands are found when using Git Bash on Windows

## Testing

## Security

- [x] No new unsafe code

### DIFF
--- a/crates/openfang-runtime/Cargo.toml
+++ b/crates/openfang-runtime/Cargo.toml
@@ -32,6 +32,7 @@ regex-lite = { workspace = true }
 rusqlite = { workspace = true }
 tokio-tungstenite = "0.24"
 shlex = "1"
+encoding_rs = "0.8"
 
 [dev-dependencies]
 tokio-test = { workspace = true }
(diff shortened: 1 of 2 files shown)


jaberjaber23 (changes requested):
Security regression in the safe path. The original code intentionally runs commands without shell interpretation (direct argv). This PR wraps ALL safe-path commands in `cmd /C` on Windows, which reintroduces shell interpretation (pipes, redirects, & all become live). This defeats the entire purpose of the safe path.

Other issues:
- `Box::leak` in `get_git_sh_path()` permanently leaks memory on every call. Use `OnceLock` or `tokio::sync::OnceCell` to compute and cache once.
- Missing `std::os::windows::process::CommandExt` import for `creation_flags(CREATE_NO_WINDOW)` — won't compile on Windows.
- PATH uses Unix colon separators but falls back to cmd.exe in some cases where semicolons are needed.

The `cmd /C` wrapping must be reverted from the safe path. The memory leak must be fixed. The missing import must be added.

jaberjaber23:
Reviewed. The Git detection logic is useful but wrapping direct-exec with cmd /C defeats the allowlist security model (any command becomes executable through cmd). PATH separator uses : instead of ; on Windows. The where.exe output is trusted without path validation. Needs architectural rethink for the Windows direct-exec path.

Jengro777:

jaberjaber23:

PR #753, opened 2026-03-19
Replace SHA256 password hashing with Argon2id for dashboard auth

RamXX:
## Summary

Dashboard password hashing used plain SHA256 without a salt. SHA256 is a general-purpose hash, not a password hash: it is fast to compute and vulnerable to rainbow tables and GPU-accelerated brute force. This replaces it with Argon2id, the OWASP-recommended password hashing algorithm, using the PHC string format with random per-hash salts.

This is a breaking change for users who have dashboard auth enabled (`[auth] enabled = true`). Existing SHA256 hashes in `config.toml` will no longer be accepted. To upgrade, run `openfang auth hash-password`, paste the new hash into `config.toml`, and restart the daemon. Dashboard auth is opt-in (disabled by default), so most users are unaffected.

No new dependencies are introduced. Both `argon2` and `rand` were already in the workspace (used by `openfang-extensions` for credential vault key derivation); this PR adds them to `openfang-api`'s Cargo.toml.

## Changes

- `hash_password()` now produces Argon2id PHC strings with random salts
- `verify_password()` only accepts Argon2id hashes (rejects legacy SHA256)
- Updated `AuthConfig::password_hash` doc comment to reflect the new format
- Added tests: Argon2id output format, unique salts, SHA256 rejection, malformed hash rejection, garbage input rejection
- Added `[auth]` section to `docs/configuration.md` (was previously undocumented)
- Updated `docs/troubleshooting.md` FAQ to reference built-in dashboard auth instead of reverse proxy workaround
- Added breaking change entry to `CHANGELOG.md`

## Testing

## Security

- [x] No new unsafe code

### DIFF
--- a/crates/openfang-api/src/server.rs
+++ b/crates/openfang-api/src/server.rs
@@ -104,6 +104,15 @@ pub async fn build_router(
             .allow_headers(tower_http::cors::Any)
     };
 
+    // Warn if dashboard auth is enabled but the password hash is not Argon2id.
+    let ph = &state.kernel.config.auth.password_hash;
+    if state.kernel.config.auth.enabled && !ph.is_empty() && !ph.starts_with("$argon2") {
+        tracing::warn!(
+            "Dashboard auth password_hash is not in Argon2id format. \
+             Login will fail. Regenerate with: openfang auth hash-password"
+        );
+    }
+
     // Trim whitespace so `api_key = ""` or `api_key = "  "` both disable auth.
     let api_key = state.kernel.config.api_key.trim().to_string();
     let auth_state = crate::middleware::AuthState {
--- a/crates/openfang-cli/src/main.rs
+++ b/crates/openfang-cli/src/main.rs
@@ -237,6 +237,9 @@ enum Commands {
         #[arg(long)]
         json: bool,
     },
+    /// Dashboard authentication [*].
+    #[command(subcommand)]
+    Auth(AuthCommands),
     /// Security tools and audit trail [*].
     #[command(subcommand)]
     Security(SecurityCommands),
@@ -680,6 +683,12 @@ enum CronCommands {
 }
 
 #[derive(Subcommand)]
+enum AuthCommands {
+    /// Generate an Argon2id password hash for dashboard authentication.
+    HashPassword,
+}
+
+#[derive(Subcommand)]
 enum SecurityCommands {
     /// Show security status summary.
     Status {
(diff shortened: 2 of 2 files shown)


RamXX:
Thank you for the thorough review. Both concerns are addressed in the latest push (e1bd2f8):

**1. `openfang auth hash-password` CLI command**

Added as `Commands::Auth(AuthCommands::HashPassword)` in the CLI. It prompts for a password with confirmation, generates the Argon2id hash, and prints a ready-to-paste `[auth]` config snippet:

```
  >> Argon2id hash generated. Add this to your config.toml:

  [auth]
  enabled = true
  password_hash = "$argon2id$v=19$m=19456,t=2,p=1$..."

  hint: Restart the daemon after updating config.toml
```

This matches what the docs reference, so users have a clear path from upgrade to working auth.

**2. Startup warning for legacy hashes**

Added a `tracing::warn` in `build_router` that fires when `auth.enabled = true` and `password_hash` is present but not in Argon2id format:

```
WARN: Dashboard auth password_hash is not in Argon2id format.
      Login will fail. Regenerate with: openfang auth hash-password
```

This covers the case where a user upgrades the binary but hasn't regenerated their hash yet. Instead of silent login failure, they get a clear log message pointing to the fix.

Happy to make any other changes you think are needed.

jaberjaber23:
Reviewed and approved. Correct Argon2id implementation with proper salting and good tests. Has merge conflicts with recent changes on main. Please rebase and we will merge immediately.

RamXX:
Rebased onto main. There were merge conflicts in `Cargo.lock` (regenerated) and I had to add the new `budget_config` field to the test `AppState` constructors that were introduced since my last push.

Worth noting: `cargo clippy --workspace` and `cargo test --workspace` both have pre-existing failures on main right now. `openfang-memory` has some `unwrap_or_default` lint violations that block clippy for any crate that depends on it, and `parse_researcher_hand` asserts a stale `max_iterations` value. Neither is related to this PR. Our changes compile and the auth tests pass.

jaberjaber23:
Reviewed and approved. Has merge conflicts with recent changes on main. Please rebase and we will merge immediately.

RamXX:
Rebased onto main (`9fef6d6`). Cargo.lock regenerated, no other conflicts. The third commit from last rebase was dropped (already upstream). Note: `openfang-runtime` has a pre-existing build failure on main (`StreamableHttpClientTransportConfig` is non-exhaustive), not related to this PR.

PR #760, opened 2026-03-20
fix(matrix): prevent echo loop by using /whoami user_id for self-message filtering

### LINKED ISSUE #757: Matrix bot stuck in loop
### Description

As soon as the Matrix user receives a message, the bot is thrown into a endless loop.

### Expected Behavior

The bot should answer the question and move on.

### Steps to Reproduce

1. Add bot to Matrix channel
2. Send message

### OpenFang Version

0.5.1

### Operating System

Linux (x86_64)

### Logs / Screenshots

taylor:
Sup

mikejones:
It's Thursday, March 19, 2026, and the current time is 21:20:52 (9:20 PM) in your local timezone, which is UTC-5.
It's Thursday, March 19, 2026, and the current time is 21:20:56 (9:20 PM) in your local timezone, which is UTC-5.
It's Thursday, March 19, 2026, and the current time is 21:21:00 (9:21 PM) in your local timezone, which is UTC-5.
It's Thursday, March 19, 2026, and the current time is 21:21:03 (9:21 PM) in your local timezone, which is UTC-5.
It's Thursday, March 19, 2026, and the current time is 21:21:06 (9:21 PM) in your local timezone, which is UTC-5.
It's Thursday, March 19, 2026, and the current time is 21:21:08 (9:21 PM) in your local timezone, which is UTC-5.
It's Thursday, March 19, 2026, and the current time is 21:21:11 (9:21 PM) in your local timezone, which is UTC-5.
It's Thursday, March 19, 2026, and the current time is 21:21:14 (9:21 PM) in your local timezone, which is UTC-5.
It's Thursday, March 19, 2026, and the current time is 21:21:17 (9:21 PM) in your local timezone, which is UTC-5.
It's Thursday, March 19, 2026, and the current time is 21:21:21 (9:21 PM) in your local timezone, which is UTC-5.
It's Thursday, March 19, 2026, and the current time is 21:21:24 (9:21 PM) in your local timezone, which is UTC-5.

sliverp:
## Summary

Fixes #757 — Matrix bot gets stuck in an infinite echo loop.

## Root Cause

The Matrix adapter's echo guard compared incoming message senders against the **config-provided** `user_id` instead of the server-validated identity from `/whoami`. When these differ (wrong casing, empty config value, different homeserver), the bot fails to recognize its own messages and replies to them indefinitely.

### 1. Echo guard: use server-validated user_id (primary fix)
- `start()` now uses the `/whoami`-validated user identity for echo filtering instead of `self.user_id`
- Added a warning log when the config `user_id` diverges from the server-reported value

### 2. Fix txn_id reuse across message chunks
- Moved `uuid::Uuid::new_v4()` generation **inside** the chunk loop in `api_send_message()`
- Previously, a single `txn_id` was generated once and reused for all chunks — Matrix's PUT idempotency silently dropped all chunks after the first

### 3. Tests
- `test_matrix_adapter_empty_user_id`: empty config user_id doesn't panic on construction
- `test_sender_echo_filter_semantics`: case-sensitive exact match verification for the echo guard
- `test_txn_id_uniqueness`: UUID uniqueness across message chunks

## Testing

All existing tests pass. New tests cover the specific failure modes that caused #757.

### DIFF
--- a/crates/openfang-channels/src/matrix.rs
+++ b/crates/openfang-channels/src/matrix.rs
@@ -68,14 +68,17 @@ impl MatrixAdapter {
         room_id: &str,
         text: &str,
     ) -> Result<(), Box<dyn std::error::Error>> {
-        let txn_id = uuid::Uuid::new_v4().to_string();
-        let url = format!(
-            "{}/_matrix/client/v3/rooms/{}/send/m.room.message/{}",
-            self.homeserver_url, room_id, txn_id
-        );
-
         let chunks = crate::types::split_message(text, MAX_MESSAGE_LEN);
         for chunk in chunks {
+            // Each chunk needs its own txn_id — Matrix deduplicates PUTs by
+            // (room_id, txn_id), so reusing the same txn_id silently drops
+            // subsequent chunks.
+            let txn_id = uuid::Uuid::new_v4().to_string();
+            let url = format!(
+                "{}/_matrix/client/v3/rooms/{}/send/m.room.message/{}",
+                self.homeserver_url, room_id, txn_id
+            );
+
             let body = serde_json::json!({
                 "msgtype": "m.text",
                 "body": chunk,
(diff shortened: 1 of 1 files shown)


jaberjaber23:
Closing. The core echo-loop fix (using /whoami user_id) was already merged in #768. The txn_id chunk fix in this PR is valid and still present on main. Please submit that as a separate small PR.

Issue #771, opened 2026-03-21
[Bug] Qwen3.5-plus fails intermittently in multi-turn conversations with "invalid_parameter_error" regarding missing tool responses


felix307253927:
### Description

When using the `qwen3.5-plus` model for multi-turn conversations involving tool calls, the application occasionally crashes with a 400 Invalid Parameter error. The error message indicates that an assistant message containing `tool_calls` was not immediately followed by the corresponding tool response messages in the conversation history.

Clearing the session (resetting the conversation history) temporarily resolves the issue, suggesting a state accumulation or message formatting bug in the agent loop.

### Expected Behavior

The runtime should correctly append the tool role messages immediately after the assistant message containing tool_calls, ensuring the message list adheres to the model's strict format requirements before sending the next request.

### Steps to Reproduce

1. Initialize a new chat session with the `qwen3.5-plus` model.
2. Engage in a multi-turn conversation that triggers at least one tool call.
3. Continue the conversation for several more turns (the issue appears probabilistically, often after 3-5 turns or specific tool interaction patterns).
4. Observe the application throwing the invalid_parameter_error.
5. Workaround: Clearing the current session/history allows the conversation to proceed normally again.

### OpenFang Version

0.5.1

### Operating System

Windows

### Logs / Screenshots

[2026-03-21][12:39:48][WARN][openfang_runtime::agent_loop] LLM stream error classified: Request failed: {"error":{"code":"invalid_parameter_error","message":"<400> InternalError.Algo.InvalidParameter: An assistant message with \"tool_calls\" must be followed by tool messages responding to each \"tool... category=Format retryable=false raw=API error (400): {"error":{"code":"invalid_parameter_error","message":"<400> InternalError.Algo.InvalidParameter: An assistant message with \"tool_calls\" must be followed by tool messages responding to each \"tool_call_id\". The following tool_call_ids did not have response messages: message[3].role","param":null,"type":"invalid_request_error"},"request_id":"b6d2d0ec-deb8-982f-b314-23114b0769fb"}

felix307253927:
This issue caused by the `call id` being overwritten by an empty string. #776

felix307253927:
Another error occurred. 

model:  `minimax-2.5`

[...]

PR #776, opened 2026-03-21
fix: Empty string IDs are overwritten, leading to inconsistencies in …

felix307253927:
…certain models.

## Summary
Empty string IDs are overwritten, leading to inconsistencies in certain models.

If the ID is an empty string, it gets overwritten, causing anomalies in models like `qwen3.5-plus` and `minimax-2.5` during multi-turn tool calls.

## Testing

## Security

- [x] No new unsafe code


felix307253927:
- #771

PR #841, opened 2026-03-25
runtime: preserve screenshot tool results as multimodal input

CastleOneX:
Problem

When a tool returns a ToolResult plus an attached screenshot, the OpenAI-compatible driver currently serializes the tool result but drops the accompanying image/text blocks on the next model turn. In practice, that means agents can capture screenshots but cannot actually reason over them as visual context.

Root cause

In crates/openfang-runtime/src/drivers/openai.rs, Role::User messages containing ToolResult blocks were converted into tool messages, but any sibling image/text content in the same turn was not preserved for the next model request.

What this changes

- In crates/openfang-runtime/src/agent_loop.rs, screenshot-bearing tool results are normalized into a sanitized ToolResult plus companion live image/text blocks.
- In crates/openfang-runtime/src/drivers/openai.rs, the OpenAI-compatible driver now preserves both the tool message and the follow-up multimodal user parts.
- Adds regression coverage for inline screenshot extraction and mixed tool+image serialization.

Why this is useful upstream

- It is not Firefox-specific.
- It improves any tool workflow that returns visual evidence.
- It fixes a real multimodal runtime gap without changing text-only or tool-only flows.

Validation

- cargo test -p openfang-runtime -j 1

### DIFF
--- a/crates/openfang-runtime/src/agent_loop.rs
+++ b/crates/openfang-runtime/src/agent_loop.rs
@@ -81,6 +81,59 @@ fn append_tool_error_guidance(tool_result_blocks: &mut Vec<ContentBlock>) {
     }
 }
 
+fn parse_inline_image_data_url(data_url: &str) -> Option<(String, String)> {
+    let rest = data_url.strip_prefix("data:")?;
+    let (metadata, payload) = rest.split_once(',')?;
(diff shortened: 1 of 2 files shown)


jaberjaber23:

PR #842, opened 2026-03-25
runtime: add config switches for memory and embeddings

CastleOneX:
Problem

OpenFang did not have an explicit runtime-level way to disable memory entirely or disable embedding-backed recall independently. In practice, memory behavior could still partially activate unless provider setup failed in a particular way, which led to unnecessary embedding initialization attempts, noisy fallback paths, and avoidable external calls.

What this changes

- Adds explicit config flags in crates/openfang-types/src/config.rs:
  - memory.enabled
  - memory.embeddings_enabled
- In crates/openfang-kernel/src/kernel.rs:
  - skips embedding driver initialization when memory is disabled
  - skips embedding driver initialization when embeddings are disabled
  - passes the effective runtime flags into agent manifest metadata
- In crates/openfang-runtime/src/agent_loop.rs:
  - skips recall entirely when memory is disabled
  - disables embedding-backed recall when embeddings are disabled
  - skips remember / remember_with_embedding writes when memory is disabled
  - does this in both streaming and non-streaming loops

Why this is useful upstream

- It makes runtime behavior explicit and configurable.
- It reduces unwanted provider calls and noisy fallback behavior.
- It gives users a supported way to choose between full memory, text-only memory, and no memory.
- Defaults remain enabled, so existing behavior is preserved unless explicitly changed.

Validation

- cargo check -p openfang-runtime -p openfang-kernel -p openfang-types

### DIFF
--- a/crates/openfang-kernel/src/kernel.rs
+++ b/crates/openfang-kernel/src/kernel.rs
@@ -851,6 +851,13 @@ impl OpenFangKernel {
             Arc<dyn openfang_runtime::embedding::EmbeddingDriver + Send + Sync>,
         > = {
             use openfang_runtime::embedding::create_embedding_driver;
+            if !config.memory.enabled {
+                info!("Memory disabled in config; runtime recall/remember and embeddings are off");
+                None
+            } else if !config.memory.embeddings_enabled {
+                info!("Memory embeddings disabled in config; using text-only memory path");
+                None
+            } else {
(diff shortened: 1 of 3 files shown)


jaberjaber23:
Reviewed. The feature logic is correct but CI is failing (format check and ubuntu test). Please run cargo fmt and add unit tests for MemoryConfig defaults and deserialization with missing fields.

jaberjaber23:

PR #847, opened 2026-03-25
fix: support Ollama 'reasoning' field in streaming and non-streaming responses

### LINKED ISSUE #805: Streaming parser does not handle Ollama's reasoning field for thinking models
### Description

When using thinking/reasoning models via Ollama (e.g. Qwen 3.5), Ollama's OpenAI-compatible streaming endpoint places the model's thinking output in a field named reasoning within the delta object. The streaming parser in openai.rs (around line 1152) only checks for reasoning_content, not reasoning. As a result, the thinking output is silently dropped.

Might be similar to issue #785 

Ollama streaming chunk example:
```{"choices":[{"delta":{"role":"assistant","content":"","reasoning":"Thinking Process:\n..."}}]}```

The parser checks:
  - delta["content"] — empty for thinking models
  - delta["reasoning_content"] — not present in Ollama's response

It does not check delta["reasoning"], which is where Ollama places the output.

### Expected Behavior

The parser should also check delta["reasoning"] as a fallback, e.g.:
  ```if let Some(reasoning) = delta["reasoning_content"].as_str().or_else(|| delta["reasoning"].as_str()) {```

### Steps to Reproduce

1. Pull a thinking model in Ollama: ollama pull qwen3.5
2. Configure OpenFang with Ollama and qwen3.5:latest
3. Send a chat message
4. The thinking output is dropped — no text appears

### OpenFang Version

0.5.1

### Operating System

Linux (x86_64)

### Logs / Screenshots

Version: 0.5.1 (confirmed by reading source at crates/openfang-runtime/src/drivers/openai.rs)

sliverp:
## Problem

Ollama uses a JSON field named `reasoning` for thinking/reasoning output from models like DeepSeek-R1 and Qwen3, while most other providers (LM Studio, etc.) use `reasoning_content`. The current implementation only checks for `reasoning_content`, which means thinking content is silently dropped when using Ollama as a provider.

Reported in #805.

## Solution

Both streaming and non-streaming code paths in the OpenAI-compatible driver now check for `reasoning` as a fallback when `reasoning_content` is absent:

- **Non-streaming** (`OaiResponseMessage` struct): Added an optional `reasoning` field. The `complete()` method now uses `.reasoning_content.or(.reasoning)` to capture thinking output from either provider.
- **Streaming** (SSE delta parsing): The delta handler now checks `delta["reasoning"]` as fallback when `delta["reasoning_content"]` is null.

## Testing

- All 826 existing tests pass (`cargo test --package openfang-runtime`)
- The fix is backward-compatible: providers that use `reasoning_content` are unaffected; the new `reasoning` field is only consulted as a fallback.

Fixes #805

### DIFF
--- a/crates/openfang-runtime/src/drivers/openai.rs
+++ b/crates/openfang-runtime/src/drivers/openai.rs
@@ -254,7 +254,10 @@ struct OaiResponseMessage {
     tool_calls: Option<Vec<OaiToolCall>>,
     /// Reasoning/thinking content returned by some models (DeepSeek-R1, Qwen3, etc.)
     /// via LM Studio, Ollama, and other local inference servers.
+    /// Most providers use `reasoning_content`; Ollama uses `reasoning`.
     reasoning_content: Option<String>,
+    /// Ollama places thinking output in a field named `reasoning` (not `reasoning_content`).
+    reasoning: Option<String>,
 }
 
 #[derive(Debug, Deserialize)]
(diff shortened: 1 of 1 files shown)


jaberjaber23:
Closing. The streaming half of this fix was already implemented in v0.5.2 (commit 604e4ea, fixing issue #805). The non-streaming half is valid but this PR will produce merge conflicts. Please rebase onto current main and submit only the non-streaming fix if needed.

Bodge-IT:
Tested, fix confirmed

PR #876, opened 2026-03-27
Add per-channel system prompt injection

pbranchu:
## Summary
- Add `system_prompt_replace` field to `ChannelOverrides` with `#[serde(default = "default_true")]` to control whether channel prompts replace or augment the agent's base prompt
- Add `send_message_with_channel_prompt` method to `ChannelBridgeHandle` trait (with default fallback to `send_message`)
- Implement kernel-level `CHANNEL_PROMPTS` static DashMap for thread-safe per-turn prompt injection without threading the value through every layer
- Add `channel_system_prompt` field to `PromptContext` and inject as Section 9.05 ("Channel Instructions") in the prompt builder
- Implement on `KernelBridgeAdapter` in openfang-api; `dispatch_message` in bridge.rs calls the new method when overrides have a system_prompt
- Fix pre-existing CI issues (clippy lints, test compilation, formatting)

## Test plan
- [ ] Verify cargo fmt, clippy, and tests pass in CI
- [ ] Configure a channel with `system_prompt` override and verify it appears in the agent's system prompt as "Channel Instructions"
- [ ] Verify channels without system_prompt override are unaffected
- [ ] Verify subagents do not receive channel instructions

### DIFF
--- a/crates/openfang-channels/src/line.rs
+++ b/crates/openfang-channels/src/line.rs
@@ -108,7 +108,7 @@ impl LineAdapter {
             diff |= a ^ b;
         }
         if diff != 0 {
-            let computed = base64::engine::general_purpose::STANDARD.encode(&result);
+            let computed = base64::engine::general_purpose::STANDARD.encode(result);
             // Log first/last 4 chars of each signature for debugging without leaking full HMAC
             let comp_redacted = format!(
                 "{}...{}",


jaberjaber23:
Reviewed. Two blocking issues:

1. Race condition: the global static CHANNEL_PROMPTS DashMap is keyed only by agent_id. If two concurrent messages hit the same agent from different channels, the second insert overwrites the first prompt before it is consumed. Thread the channel prompt through the call stack instead, or key by (agent_id, session_id).

2. system_prompt_replace is declared in config but never read anywhere. The behavior is always "append" regardless of the flag value. Either implement the replace logic or remove the dead field to avoid misleading operators.

Also: no cleanup on panic/cancellation (stale prompt leak). Use a scopeguard or Drop pattern around the insert/remove.

PR #877, opened 2026-03-27
Fix silent reinforcement: recognize [SILENT] token

pbranchu:
## Summary
- Add `[SILENT]` and `[silent]` to the NO_REPLY detection conditions in both streaming and non-streaming agent loop paths
- When the agent outputs these tokens, store `[no reply needed]` in history instead of the literal `[SILENT]` text, which would otherwise reinforce silence in future turns
- Fix pre-existing CI issues (clippy lints, test compilation, formatting)

## Test plan
- [ ] Verify cargo fmt, clippy, and tests pass in CI
- [ ] Send a message where the agent should stay silent; verify `[SILENT]` is not stored literally in session history
- [ ] Verify `NO_REPLY` token still works as before
- [ ] Verify `[[silent]]` directive parsing still works as before

### DIFF
--- a/crates/openfang-channels/src/line.rs
+++ b/crates/openfang-channels/src/line.rs
@@ -108,7 +108,7 @@ impl LineAdapter {
             diff |= a ^ b;
         }
         if diff != 0 {
-            let computed = base64::engine::general_purpose::STANDARD.encode(&result);
+            let computed = base64::engine::general_purpose::STANDARD.encode(result);
             // Log first/last 4 chars of each signature for debugging without leaking full HMAC
             let comp_redacted = format!(
                 "{}...{}",


jaberjaber23:
Reviewed. Three issues before merge:

1. Case sensitivity gap: the check matches [SILENT] and [silent] but misses [Silent] or other mixed-case variants that LLMs produce. Use .to_lowercase() for the comparison.

2. No unit tests for the new [SILENT] detection. Add tests asserting that [SILENT], [silent], and [Silent] all route to the no-reply path and store "[no reply needed]" in session history.

3. The openssl-src addition in Cargo.lock is unexplained and unrelated to this feature. Please explain or remove from this PR.

PR #879, opened 2026-03-27
Add qwen3.5-flash, qwen3.5-plus, qwen3-max model

shilkazx:
## Testing

- [ ] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [ ] `cargo test --workspace` passes
- [ ] Live integration tested (if applicable)

## Security

- [ ] No new unsafe code
- [ ] No secrets or API keys in diff
- [ ] User input validated at boundaries

PR #899, opened 2026-03-28
fix(model-catalog): correct qwen3.5-flash, qwen3.5-plus, qwen3-max entries

dvrd:
## Summary

Fixes 11 data errors in the three new Qwen model entries added in #879:

- **qwen3.5-flash**: `output_cost_per_m` 2.00→0.60 (3.3× overcharge), `context_window` 991K→1M, removed self-alias, `supports_vision` true→false, `max_output_tokens` 64K→8192
- **qwen3.5-plus**: `output_cost_per_m` 4.80→3.20 (published Alibaba rate), `context_window` 991K→1M, `supports_vision` true→false, `max_output_tokens` 64K→8192
- **qwen3-max**: `tier` Smart→Frontier (it's Alibaba's top-tier managed endpoint), `context_window` 252K→1M, `supports_vision` true→false, `max_output_tokens` 64K→8192
- **test**: replace hardcoded provider count assertion (`== 41`) with named-provider checks

All three models are text-only (vision belongs to the separate `qwen3.5-vl` family). `max_output_tokens: 64_000` does not appear in Alibaba Cloud docs for any of these models — all Qwen models in the catalog use 8_192.

## Test plan

- [x] `cargo test -p openfang-runtime` — 846 passed, 0 failed
- [x] `cargo build --workspace --lib` — clean build

### DIFF
--- a/crates/openfang-runtime/src/model_catalog.rs
+++ b/crates/openfang-runtime/src/model_catalog.rs
@@ -2960,26 +2960,26 @@ fn builtin_models() -> Vec<ModelCatalogEntry> {
             display_name: "Qwen3.5 Flash".into(),
             provider: "qwen".into(),
             tier: ModelTier::Fast,
-            context_window: 991_000,
-            max_output_tokens: 64_000,
+            context_window: 1_000_000,
+            max_output_tokens: 8_192,
             input_cost_per_m: 0.20,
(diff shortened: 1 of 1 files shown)


dvrd:
Closing — submitting against our fork instead.

PR #906, opened 2026-03-29
fix: select crypto provider (#903)

### LINKED ISSUE #903: rustls 0.23 CryptoProvider panic on Mattermost wss:// connection
## Problem

Any TLS connection (specifically the Mattermost channel adapter's wss:// WebSocket) causes a tokio worker thread panic:

```
thread 'tokio-rt-worker' panicked at rustls-0.23.37/src/crypto/mod.rs:249:14:

Could not automatically determine the process-level CryptoProvider from Rustls crate features.
Call CryptoProvider::install_default() before this point to select a provider manually,
or make sure exactly one of the 'aws-lc-rs' and 'ring' features is enabled.
```

After enough panics accumulate (~50), the process exits. This happens on every startup when `[channels.mattermost]` is configured with an `https://` server URL.

## Environment

- v0.5.4 and v0.5.5 Linux x86_64 binaries (both affected)
- Docker on Ubuntu 22.04
- Config has `[channels.mattermost]` with `server_url = "https://chat.example.com"`

## Root cause

Cargo.toml uses `rustls-tls` feature for reqwest and `rustls-tls-native-roots` for tokio-tungstenite, but rustls 0.23+ requires an explicit CryptoProvider to be installed before any TLS operations. The binary doesn't call `CryptoProvider::install_default()` at startup.

## Suggested fix

Add this early in main(), before any async runtime or TLS connections:

```rust
rustls::crypto::ring::default_provider().install_default().ok();
```

Or ensure the `ring` crate feature is enabled for rustls in Cargo.toml so it auto-installs.

## Workaround

Remove `[channels.mattermost]` from config.toml entirely. This disables native Mattermost DM-to-agent chat but avoids the panic. Slash commands via an external bridge still work.

hello-world-bfree:
## Summary

Install the rustls CryptoProvider (`ring`) at kernel boot to fix a panic when channel adapters (e.g. Discord) open TLS connections. Rustls 0.23+ can't auto-select a provider when both `ring` and `aws-lc-rs` are in the dependency tree. Resolves https://github.com/RightNow-AI/openfang/issues/903

> [!NOTE]
> There's also an issue with the Discord Gateway protocol. At the moment the client does not send a heartbeat. This adds that in.

## Changes

- Added `rustls` (with `ring` feature) as a workspace dependency       - Added `rustls` to `openfang-kernel` crate                           - Call `install_default()` as the first line of `boot_with_config()`, before any `reqwest::Client` construction                                                                                                    

## Testing

## Security

- [x] No new unsafe code                                               

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4179,6 +4179,7 @@ dependencies = [
  "openfang-wire",
  "rand 0.8.5",
  "reqwest 0.12.28",
+ "rustls 0.23.37",
  "serde",
  "serde_json",
  "subtle",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,6 +63,7 @@ clap_complete = "4"
 
 # HTTP client (for LLM drivers)
 reqwest = { version = "0.12", default-features = false, features = ["json", "stream", "multipart", "rustls-tls", "gzip", "deflate", "brotli"] }
+rustls = { version = "0.23", default-features = false, features = ["ring"] }
 
 # Async trait
 async-trait = "0.1"
--- a/crates/openfang-kernel/Cargo.toml
+++ b/crates/openfang-kernel/Cargo.toml
@@ -33,6 +33,7 @@ subtle = { workspace = true }
 rand = { workspace = true }
 hex = { workspace = true }
 reqwest = { workspace = true }
+rustls = { workspace = true }
 cron = "0.15"
 zeroize = { workspace = true }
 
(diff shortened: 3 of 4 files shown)


jaberjaber23:
Reviewed. Two issues: (1) The Discord heartbeat fix is unrelated to the crypto provider fix. Split into two PRs. (2) The install_default() Result is silently discarded with let _ =. At minimum add a debug log.

jaberjaber23:
Closing — we asked to split this PR (crypto provider install is a separate concern from the Discord heartbeat fix) and received no response. The rustls crypto provider fix is now partially addressed by #942. If the Discord heartbeat fix is still needed, please resubmit it as a focused PR.

Issue #926, opened 2026-03-31
Fail to compile: cannot create non-exhaustive struct using struct expression


nxtreaming:
### Description

 Compiling openfang-runtime v0.5.5 (/openfang/crates/openfang-runtime)
error[E0639]: cannot create non-exhaustive struct using struct expression
   --> crates/openfang-runtime/src/mcp.rs:310:22
    |
310 |           let config = StreamableHttpClientTransportConfig {
    |  ______________________^
311 | |             uri: Arc::from(url),
312 | |             custom_headers,
313 | |             ..Default::default()
314 | |         };
    | |_________^

For more information about this error, try `rustc --explain E0639`.
error: could not compile `openfang-runtime` (lib) due to 1 previous error

### Expected Behavior

Compile success.

### Steps to Reproduce

cargo build --profilie release -p openfang-cli

### OpenFang Version

0.5.7

### Operating System

macOS (Apple Silicon)

snowball77:
The same issue arises on docker compose build

jaberjaber23:
Fixed in a previous release. mcp.rs now uses the rmcp builder API: StreamableHttpClientTransportConfig::with_uri(url).custom_headers(custom_headers). Verified the code compiles clean on current main.

PR #927, opened 2026-03-31
fix(runtime): handle non-exhaustive MCP HTTP config

varpress:
fix: #926 
1. Fail to compile: cannot create non-exhaustive struct using struct expression


Jengro777:

jaberjaber23:
Closing as duplicate of #986, which uses the canonical `with_uri(...).custom_headers(...)` builder chain end-to-end (no public field assignment). Thanks @varpress for the independent fix.

PR #964, opened 2026-04-02
Fix WebSocket auth rejecting API keys with URL-special characters

### LINKED ISSUE #962: WebSocket auth rejects valid API keys containing URL-special characters — chat silently falls back to HTTP (no streaming)
### Description

The WebSocket upgrade handler in `crates/openfang-api/src/ws.rs` authenticates via a `?token=` query parameter for browser clients (browsers cannot set custom headers on WebSocket connections). The handler extracts the token value with a raw `split('&')` / `strip_prefix("token=")` approach and compares it directly against the configured API key.

The bug: browsers call `encodeURIComponent()` on the token before appending it to the WebSocket URL (implemented in `static/js/api.js`). Characters common in base64-derived keys — `+`, `/`, `=` — are percent-encoded to `%2B`, `%2F`, `%3D`. The server receives the encoded string and compares it character-for-character against the raw key. They never match, so every WebSocket upgrade returns 401.

The fallback path in `chat.js` silently switches to HTTP polling mode with a toast notification ("Using HTTP mode (no streaming)"). The user sees responses arriving as complete messages rather than streaming tokens, with no clear indication that the root cause is an API key character encoding issue.

**Affected users**: any installation where the API key contains `+`, `/`, or `=` — this includes virtually all base64-derived keys (e.g., keys generated with `openssl rand -base64 32`). Keys using only alphanumeric characters and `-` / `_` are not affected.

**Workaround**: generate an API key that contains only URL-safe characters (alphanumeric, `-`, `_`). Example: `openssl rand -hex 32`.

### Expected Behavior

The chat dashboard uses WebSocket for all agent communication regardless of API key content. Streaming tokens appear in real time. User does not have to wait until entire response is formed to see its contents.

### Steps to Reproduce

1. Configure an API key containing `+`, `/`, or `=` (any base64-derived key):
   ```bash
   # Example key that triggers the bug:
   openssl rand -base64 32
   # → Zelrgr4sRcPzu4VNFcgbdykL02nT+h73l1mH3jlvjgZL9+QvlJ7uiBdZCs/gPwKz
   ```
2. Set in environment: `OPENFANG_API_KEY=<key above>` with `api_key = ""` in `config.toml`
3. Open the dashboard and navigate to any agent's chat view
4. Send a message

**Actual**: toast notification "Using HTTP mode (no streaming)" appears; responses arrive as complete blocks; WebSocket connection never establishes.

**Expected**: WebSocket connects immediately on agent selection; responses stream token by token.

## Debugging confirmation

Server-side the endpoint is reachable. A raw WebSocket upgrade with the literal (non-encoded) key succeeds (101 Switching Protocols). A request with `encodeURIComponent`-encoded token returns 401.

### OpenFang Version

0.5.7

### Operating System

Linux (x86_64)

### Logs / Screenshots

_No response_

chrisyoung2005:
## What

Fixes #962.

The WebSocket upgrade handler authenticates browser clients via a `?token=` query parameter (browsers cannot set custom headers on WebSocket connections). The handler was extracting the token with `split('&')` / `strip_prefix("token=")` and comparing it directly against the configured API key using constant-time comparison.

**The problem**: browsers call `encodeURIComponent()` on the token before appending it to the WebSocket URL (`api.js:226`). Characters common in base64-derived keys — `+`, `/`, `=` — are percent-encoded to `%2B`, `%2F`, `%3D`. The server received the encoded string and compared it character-for-character against the raw key. They never match, so every WebSocket upgrade returned 401.

The result: the fallback in `chat.js` silently switched to HTTP polling mode. Users with base64-derived API keys (e.g. generated with `openssl rand -base64 32`) could never get streaming — without any clear error pointing to the root cause.

## How

Switch to `url::form_urlencoded::parse()` to decode the query string before comparison. This is an existing workspace dependency (`url = "2"` in root `Cargo.toml`) — no new transitive dependencies are added.

**Before** (`ws.rs`):
```rust
let query_auth = uri
    .query()
    .and_then(|q| q.split('&').find_map(|pair| pair.strip_prefix("token=")))
    .map(|token| ct_eq(token, api_key))
    .unwrap_or(false);
```

**After**:
```rust
let query_auth = uri
    .query()
    .and_then(|q| {
        url::form_urlencoded::parse(q.as_bytes()).find_map(|(k, v)| {
            if k == "token" {
                Some(v.into_owned())
            } else {
                None
            }
        })
    })
    .map(|token| ct_eq(&token, api_key))
    .unwrap_or(false);
```

## Verification

- `cargo clippy -p openfang-api --all-targets -- -D warnings`: zero warnings
- `cargo test -p openfang-api`: 7/7 tests pass
- Full workspace build is currently blocked by the pre-existing mcp.rs build failure (tracked in #926, fix open in #927)

## Workaround (until merged)

Generate an API key using only URL-safe characters:
```bash
openssl rand -hex 32
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4014,6 +4014,7 @@ dependencies = [
  "tower",
  "tower-http",
  "tracing",
+ "url",
  "uuid",
 ]
 
--- a/crates/openfang-api/Cargo.toml
+++ b/crates/openfang-api/Cargo.toml
@@ -32,6 +32,7 @@ futures = { workspace = true }
 governor = { workspace = true }
 tokio-stream = { workspace = true }
 subtle = { workspace = true }
+url = { workspace = true }
 base64 = { workspace = true }
 sha2 = { workspace = true }
 hmac = { workspace = true }
--- a/crates/openfang-api/src/ws.rs
+++ b/crates/openfang-api/src/ws.rs
@@ -168,8 +168,16 @@ pub async fn agent_ws(
 
         let query_auth = uri
             .query()
-            .and_then(|q| q.split('&').find_map(|pair| pair.strip_prefix("token=")))
-            .map(|token| ct_eq(token, api_key))
+            .and_then(|q| {
+                url::form_urlencoded::parse(q.as_bytes()).find_map(|(k, v)| {
+                    if k == "token" {
+                        Some(v.into_owned())
+                    } else {
+                        None
+                    }
+                })
+            })
+            .map(|token| ct_eq(&token, api_key))
             .unwrap_or(false);
 
         if !header_auth && !query_auth {


jaberjaber23:
Closing. This was already fixed on main. Thanks for the contribution.

PR #977, opened 2026-04-03
Add runtime hand loading from $OPENFANG_HOME/hands/ directory

pbranchu:
## Problem

All hand definitions must currently be compiled into the binary as bundled hands. There is no way to ship deployment-specific hands (e.g. household integrations, private API hands) without forking the source and recompiling.

## Solution

Scan `$OPENFANG_HOME/hands/` at startup and load any TOML files found there as hand definitions, alongside bundled hands.

### Two layouts supported

```
$OPENFANG_HOME/hands/
  my-hand.toml              # flat file — hand TOML, no skill content
  another-hand/
    HAND.toml               # subdirectory — hand TOML + optional SKILL.md
    SKILL.md
```

### `auto_activate` field

A new optional field `auto_activate = true` on a hand definition causes the kernel to activate it automatically on first boot (when it isn't already in `hand_state.json`). On subsequent restarts the hand is restored from `hand_state.json` like any other hand — no scripts or init containers needed.

```toml
id = "my-integration"
name = "My Integration"
auto_activate = true   # activate on first boot, persist via hand_state.json
...
```

## Changes

- **`HandDefinition`**: add `auto_activate: bool` with `#[serde(default = false)]` — fully backwards compatible, existing hands unaffected
- **`HandRegistry::load_from_directory()`**: scans a directory, upserts hand definitions from flat `.toml` files or `*/HAND.toml` subdirectories
- **Kernel boot**: calls `load_from_directory($OPENFANG_HOME/hands/)` after `load_bundled()`
- **`start_background_agents()`**: auto-activates hands with `auto_activate = true` that aren't already in persisted state

## Test plan

- [ ] Create a `hands/test-hand.toml` in `$OPENFANG_HOME/hands/`, restart — hand appears in `/api/hands`
- [ ] Add `auto_activate = true` — hand activates automatically on restart without API call
- [ ] After first activation, `hand_state.json` persists it — no double-activation on next restart
- [ ] Bundled hands unaffected (`auto_activate` defaults to `false`)
- [ ] Invalid TOML in hands dir logs a warning and continues loading others

### DIFF
--- a/crates/openfang-hands/src/registry.rs
+++ b/crates/openfang-hands/src/registry.rs
@@ -105,6 +105,69 @@ impl HandRegistry {
             .collect()
     }
 
+    /// Load hand definitions from a directory at runtime.
+    ///
+    /// Supports two layouts:
+    /// - `hands/*.toml`      — flat TOML file per hand (no SKILL.md)
+    /// - `hands/*/HAND.toml` — subdirectory with HAND.toml and optional SKILL.md
+    ///
+    /// Returns the count of hand definitions loaded.
+    pub fn load_from_directory(&self, dir: &std::path::Path) -> usize {
+        if !dir.exists() {
+            return 0;
+        }
+        let entries = match std::fs::read_dir(dir) {
+            Ok(e) => e,
+            Err(e) => {
+                warn!(path = %dir.display(), error = %e, "Failed to read hands directory");
+                return 0;
+            }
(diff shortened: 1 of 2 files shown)


pbranchu:
Closing — the existing hand persistence mechanism (hand_state.json) is sufficient. No kernel changes needed.

PR #985, opened 2026-04-05
fix: use builder API for StreamableHttpClientTransportConfig

marcoziti:
`StreamableHttpClientTransportConfig` is `#[non_exhaustive]`, so it cannot be constructed via a struct literal. Replace with the builder pattern using `with_uri()` and `custom_headers()`, and drop the now- unused `std::sync::Arc` import.

Fixes `cargo install --path crates/openfang-cli` compile error E0639.

## Testing

## Security

- [x] No new unsafe code


marcoziti:
Fix available already (e.g. PR968).

PR #1002, opened 2026-04-07
fix(chat): prevent duplicate message keys on first chat welcome message

lc-soft:
## Summary

In the first chat session, the welcome message used a local id counter, while new user messages used the global message counter.

This could generate duplicate ids and cause key collisions in message list rendering, so the chat UI did not update correctly: both the new user message and assistant reply could be missing from the visible list until a page refresh.

screenshot-20260407-151254

## Changes

Updated welcome message id generation to use the global message counter, preventing duplicate message ids.

## Testing

## Security

- [x] No new unsafe code


jaberjaber23 (approved):
LGTM. Local counter shadowed the global msgId, causing duplicate keys and rendering bugs on first chat.

PR #1022, opened 2026-04-09
fix(analytics): correct bar style attribute binding

lc-soft:
## Summary

Fixed an issue in the **analysis** page where `:style` binding with a string value would overwrite the initial `style` attribute instead of merging with it. By switching to an object-based binding (e.g., `:style="{ width: barWidth(m) }" `), the styles are now correctly applied and merged as expected.

## Changes

Before:

screenshot-20260409-191624

After:

screenshot-20260409-191059

## Testing

## Security

- [x] No new unsafe code


jaberjaber23 (approved):
LGTM. Object syntax is cleaner and avoids string concatenation edge cases.

PR #1025, opened 2026-04-09
feat: Add OAuth provider support (Codex, Gemini, Qwen, MiniMax)

AlexZander85:
## Summary
- Adds OAuth2 device code flow for OpenAI Codex, Gemini, Google, Qwen, and MiniMax
- Fixes pre-existing mcp.rs compilation error (StreamableHttpClientTransportConfig)
- Adds i18n support (Russian + English)

## Changes
- oauth_providers.rs (638 lines) — OAuth logic for 4 providers
- API routes for OAuth start/poll flow
- Dashboard UI with OAuth login buttons

## Testing
- cargo build --workspace --lib — compiles successfully
- Live test: curl http://127.0.0.1:4200/api/providers/openai-codex/oauth/start


jaberjaber23 (changes requested):
## Security Audit — REQUEST CHANGES

### CRITICAL: Weak PKCE code verifier generation (oauth_providers.rs)

The `generate_pkce()` function uses `SystemTime::now().as_nanos()` XOR'd with a static pattern as the PKCE code verifier. This is **not cryptographically random** — it is predictable/brute-forceable:

```rust
let now = SystemTime::now()
    .duration_since(SystemTime::UNIX_EPOCH)
    .unwrap_or_default()
    .as_nanos() as u64;
for i in 0..32 {
    bytes[i] = ((now >> i) & 0xFF) as u8 ^ ((i as u8) * 0x5A);
}
```

The `now` value has maybe ~30 bits of effective entropy (nanosecond counter within a second, which is ~10^9 ≈ 30 bits). The XOR with `i * 0x5A` is a constant and adds zero entropy. An attacker who knows approximately when the flow started can brute-force the verifier. **PKCE verifiers MUST use a CSPRNG** (e.g., `rand::rngs::OsRng`, or `getrandom` crate).

Same issue affects `generate_state()` — the OAuth state parameter must also be cryptographically random to prevent CSRF.

**Fix:** Replace with `getrandom::getrandom(&mut bytes).expect("CSPRNG failed")` or `rand::rngs::OsRng.fill_bytes(&mut bytes)`.

### Other findings (non-blocking but should fix):

1. **PR overlaps with #1015** — This PR includes the full i18n framework (en.json, ru.json, i18n.js, and all JS page modifications). PR #1015 adds the same files. One of these will conflict. Coordinate which lands first.

2. **Qwen OAuth reads from `~/.qwen/oauth_creds.json`** — This is a file-based "OAuth" that just reads tokens from disk. Not a real OAuth flow. The poll endpoint will always return the same static tokens. This is fine for now but misleading as "OAuth".

3. **MiniMax OAuth is a stub** — `minimax_start_oauth_flow()` returns `Err(...)` always. The UI will show a "Login with MiniMax" button that will always fail. Either implement or remove.

4. **No rate limiting on OAuth start endpoints** — The `/oauth/start` endpoints create new device flows with no rate limiting. An attacker could spam these to exhaust device codes.

5. **Token returned in poll response JSON** — The access_token is returned in the HTTP response body from the poll endpoint. This is technically correct for the device flow pattern (the dashboard is local), but ensure the dashboard is not exposed to the internet without auth.

6. **No hardcoded secrets found** — Client IDs are public (as expected for device flow). All OAuth URLs point to legitimate provider endpoints (auth.openai.com, accounts.google.com, chat.qwen.ai, api.minimax.io). No suspicious redirect URLs.

7. **Line count justified** — The 3212 additions are legitimate: ~638 lines oauth_providers.rs, ~1215 lines i18n JSON files (en+ru), ~230 lines i18n.js, ~415 lines route handlers, and ~700 lines UI/JS changes. No auto-generated padding.

**Verdict: Fix the PKCE/state generation to use a CSPRNG, then this is ready to merge.**

jaberjaber23:
PKCE code verifier uses SystemTime as sole entropy source instead of a CSPRNG. This is a critical security defect for an authentication flow. Please regenerate PKCE verifiers and state params using getrandom or OsRng and resubmit.

Issue #794, opened 2026-03-23
[BUG]Security / policy bypass: destructive Remove-Item runs via powershell -Command while del is blocked; allowlist only checks outer command


1-2-3:
## Description

When asking the agent to delete files under d:\temp, OpenFang first runs a cmd.exe-style delete that is correctly blocked by execution policy. It then runs a PowerShell one-liner that deletes the same content successfully, even though Remove-Item is not listed in allowed_commands.

This suggests the allowlist is applied to the invoked executable (e.g. powershell) but not to nested commands passed as arguments (-Command), which allows destructive operations that would be blocked if expressed as a direct shell builtin or a disallowed command name.

## Environment

OpenFang version: v0.5.1
OS: Windows 11 25H2
Configuration
Relevant config.toml excerpts:

[approval]
require_approval = ["file_delete"]
[exec_policy]
allowed_commands = [
    "ls",
    "cd",
    "grep",
    "npx",
    "npm",
    "sqlplus",
    "python",
    "print",
    "pip",
    "EXIT",
    "exit",
    "echo",
    "find",
    "sql",
    "EOF",
    "tnsping",
    "lsnrctl",
    "dir",
    "obsidian-cli",
    "obsidian",
    "powershell",
    "df",
    "wb",
    "print",
    "uvx",
    "uv",
    "netstat",
    "ping",
    "timeout",
    "sleep",
    "chrome",
    "chrome.exe",
    "pdflatex"
]
mode = "allowlist"
Note: Remove-Item is not in allowed_commands; powershell is.

## Expected behavior

One of the following (product decision):

A) Any path that results in file_delete-class behavior should require approval when [approval] require_approval = ["file_delete"] is set, regardless of whether the agent uses cmd, powershell, or another wrapper; or
B) In allowlist mode, invocations like powershell -Command "..." should be parsed or restricted so that disallowed cmdlets / operations (e.g. Remove-Item -Recurse -Force) cannot run unless explicitly allowed; or
C) Document clearly that allowing powershell (or cmd) effectively bypasses granular allowed_commands for anything expressible inside -Command, and recommend not whitelisting them if strict command control is required.

## Steps to reproduce

Use the above exec_policy (allowlist including powershell, excluding Remove-Item / del as standalone entries if that matters for your design).
Send a prompt similar to:
Please delete all files in the d:\temp folder.
(Use a test directory with disposable files.)
Observe the sequence of commands OpenFang executes.
Observed behavior
[...]

PR #1030, opened 2026-04-10
fix(security): CSPRNG for PKCE/state + OAuth improvements

AlexZander85:
## Summary
Fixes critical security vulnerabilities found in security audit of #1025.

## Security Fixes (CRITICAL)
- **PKCE code verifier**: Replace SystemTime-based pseudo-random with OsRng (256-bit CSPRNG)
- **OAuth state parameter**: Replace SystemTime with OsRng (128-bit CSPRNG)
- Both now use \
and::rngs::OsRng.fill_bytes()\ per RFC 7636 and RFC 6749

## Changes
- \generate_pkce()\: 32 bytes from OsRng → base64url (43 chars)
- \generate_state()\: 16 bytes from OsRng → base64url
- Added uniqueness tests to verify CSPRNG behavior
- Clarified MiniMax stub error messages

## Testing
\\\ash
cargo test -p openfang-runtime -- oauth
# 6 tests passed including pkce_uniqueness and state_uniqueness
\\\

## Addresses Audit Findings
- ✅ CRITICAL: Weak PKCE code verifier generation (fixed)
- ✅ CRITICAL: Weak state parameter generation (fixed)
- ✅ MiniMax stub clarified (error messages updated)

Co-authored-by: Security Audit <security@openfang.sh>


AlexZander85:
> **⚠️ Security Audit — pre-existing vulnerabilities, not introduced by this PR**
> 
> The `cargo audit` failures are all pre-existing in `main` and not caused by this PR's changes:
> 
> | Advisory | Severity | Affected crate | Fix |
> |---|---|---|---|
> | RUSTSEC-2026-0095 | 🔴 Critical | wasmtime 41.0.4 (Winch sandbox escape) | Upgrade to ≥42.0.2 |
> | RUSTSEC-2026-0049 | Medium | rustls-webpki 0.102.8 (CRL matching) | Upgrade to ≥0.103.10 |
> | RUSTSEC-2026-0097 | Low | rand 0.7/0.8/0.9 (unsound with custom logger) | Indirect deps |
> | RUSTSEC-2025-0134 | Info | rustls-pemfile 2.2.0 | — |
> 
> The critical one (wasmtime) is a major version bump that may break the sandbox API and should be handled in a separate dedicated PR. All other CI checks (build, test, clippy, fmt) pass on all platforms.

jaberjaber23 (changes requested):
Thanks for the substantial OAuth work @AlexZander85. I went through this line-by-line and there's valuable material here, but the PR as currently structured can't land. Requesting changes — detail below.

### Blockers (must fix)

1. **The advertised CSPRNG fix is dead code.** `generate_pkce` and `generate_state` in `oauth_providers.rs` are cryptographically fine (32 B OsRng → base64url, 128-bit state, S256), but neither is ever called from the new `openai_codex_*` / `gemini_*` / `qwen_*` / `minimax_*` HTTP handlers. The flows use device-code only. Either wire PKCE/state into the real paths or remove the helpers and retitle the PR.
2. **`minimax_start_oauth_flow()` always returns `Err(...)`** but the `minimax_oauth_start` route matches on `Ok(_)` for success. The endpoint 500s unconditionally.
3. **Regression vs existing `copilot_oauth_poll` pattern.** The new poll handlers return `access_token` / `refresh_token` **directly in the JSON body** to the browser. The existing Copilot flow persists server-side and returns `{"status": "complete"}`. Please align: write to the vault / env and stop echoing tokens to the client.
4. **Security test deletion.** The shell-wrapper bypass tests in `crates/openfang-runtime/src/subprocess_sandbox.rs` that were added to guard #794 are removed in this diff. A PR branded "security" cannot delete security tests — please restore them.
5. **Provider default regressions in `drivers/mod.rs`:**
   - `github-copilot` is flipped to `ProviderDefaults::simple(..., "GITHUB_TOKEN", true)` with `key_required: true` — breaks OAuth-only Copilot users.
   - `azure` and `vertex-ai` default arms are removed; `provider_defaults("azure")` / `"vertex-ai"` will return `None`.
   Restore both.
6. **Workspace version downgraded** `0.5.9 → 0.5.7` in root `Cargo.toml`. Unless intentional (it shouldn't be — 0.5.9 is shipping), revert.
7. **CI — Security Audit: FAILURE** on this branch. Must be green before merge.

### Concerns (please address or justify)

- Module header comments claim "PKCE + device code" but only device code is implemented (see blocker 1).
- `test_pkce_uniqueness` / `test_state_uniqueness` only assert two draws differ — that's not a CSPRNG verification, it's a `ne` check. Either drop them or add real entropy/statistical tests.
- `GeminiFlowState` fields carry `#[allow(dead_code)]` — sign of unfinished integration.
- No Origin/CSRF/session binding on the new `/oauth/*` endpoints; rate limiter alone isn't replay protection.
- Does not address the B1/B2 auth-bypass in `middleware.rs` from the #1034 disclosure (the `if !api_key.is_empty()` branch that fails open when `OPENFANG_API_KEY` is unset). Whoever lands the OAuth work should coordinate with that fix.

### Recommendation

Please split this into three PRs so they can be reviewed and landed independently:

1. **OAuth provider plumbing** (Codex / Gemini / Qwen / MiniMax device flows + vault-based token storage matching Copilot). Fix the MiniMax bug, don't echo tokens, don't regress Copilot/Azure/Vertex defaults.
2. **CSPRNG wiring** — only if you wire `generate_pkce` / `generate_state` into a real authorization-code flow. Otherwise drop the helpers.
3. **UI changes** in `settings.js`.

Restore the subprocess sandbox tests, revert the version bump, and keep an eye on #1034 so this doesn't paper over the real bypass.

We'd genuinely like this to land — just not in a single 1920-line bundle with a dead-code headline fix.

gudzenkov:
@AlexZander85 
Do you have ETA on this PR series for OAuth support.
Looking forward to beta-test that once you finalize branch conflicts & security-related concerns.

jaberjaber23:
Closing in the same wave as your other PRs (#1115/#1116/#1117). Each had detailed CHANGES_REQUESTED reviews that went unanswered. If you want to rework, please submit a single coherent PR with passing CI, no duplicate splits, and the specific concerns addressed: PKCE verifier correctness, fail-close semantics, token-leak in MiniMax handler. Happy to re-review then.

PR #1035, opened 2026-04-11
feat(discord): smart auto-thread mode with race-fix and proper defaults

Hypn0sis:
## Summary

Adds a `smart` auto-thread mode for the Discord adapter and fixes two bugs introduced during its development.

### New feature: `auto_thread = "smart"`
Adds a new mode alongside the existing `true`/`false` options. When set to `smart`, the bot creates a Discord thread only when it is @mentioned — not for every message in a server channel.

```toml
[channels.discord]
auto_thread = "smart"   # "false" (default) | "true" | "smart"
```

### Bug fix: default changed from `"true"` to `"false"`
The previous default caused the bot to auto-thread **every** group message even without being mentioned. The default is now `"false"` (opt-in).

### Bug fix: thread name derived from message content
Thread names were `"Thread for {sender}"` which is meaningless. Now the thread name is the first ~100 chars of the message text (Discord's limit), with `<@mention>` prefixes stripped.

### Bug fix: eliminate race-window duplicate responses when @tagged
Discord sends `MESSAGE_UPDATE` (embed resolution) ~100–500 ms after `MESSAGE_CREATE`. The dedup guard only inserted the message-id into `threaded_message_ids` *after* `create_thread()` completed (async HTTP, up to ~2 s), so `MESSAGE_UPDATE` could slip through and produce a second free-channel response alongside the thread reply.

Fix: mark the message-id as seen **immediately** when `MESSAGE_CREATE` is forwarded to the stream — before the bridge picks it up.

### Refactor: auto-thread creation moved after policy guards
Thread is now created only after DM/group policy, RBAC, and rate-limit checks all pass — prevents creating empty/orphan threads for rejected messages.


Hypn0sis:
Closing in favour of #1040 (Discord fix, 3 files) and #1041 (security deps). Split to keep each PR reviewable.

PR #1040, opened 2026-04-11
fix(discord): smart auto-thread – content-based name, default off, no duplicate responses

Hypn0sis:
## Problem

Three bugs in the Discord adapter:

1. **Every message got threaded** regardless of `auto_thread` setting, because `default_auto_thread()` returned `"true"`
2. **Thread name was always the sender's display name** instead of the message content
3. **Duplicate responses when tagged** – Discord fires `MESSAGE_UPDATE` ~100ms after `MESSAGE_CREATE` (embed resolution); if `create_thread()` takes >100ms the dedup set isn't populated yet, so a second free-channel response was dispatched

## Fix

- `config.rs`: change `default_auto_thread()` from `"true"` → `"false"` (opt-in)
- `discord.rs`: add `thread_name_from_message()` – strips `<@mention>` prefix tokens, truncates to 100 chars (Discord limit); insert message ID into dedup set synchronously on `MESSAGE_CREATE` before async thread creation
- `bridge.rs`: initialise `effective_thread_id` from `message.thread_id` unconditionally; move auto-thread creation after all policy guards; fix `api_send_thread_message` to POST to `channels/{thread_id}/messages`

## Diff size

3 files, ~350 lines – intentionally no formatting or dependency changes.


Hypn0sis:
⚠️ **Note for reviewers**: the Security Audit CI job will fail on this PR due to pre-existing CVEs in `wasmtime 41` and `rumqttc 0.24` (RUSTSEC-2026-0049, 0085–0096). Those are fixed in the companion PR #1041 — the two PRs are intentionally split to keep each one reviewable. Merging #1041 first will make this PR's CI go green.

Hypn0sis:
reopened with newer code

PR #1045, opened 2026-04-12
fix: system prompt and identity handling, and config form hydration

dongtran16092006:
## Summary
Fix issues where `system_prompt` and identity fields were not properly returned from the agent detail API or correctly populated in the UI when editing an agent.

## Changes
* Expose `system_prompt` and full `identity` fields in the agent detail API response (`routes.rs`)
* Ensure agent detail modal fetches the full agent payload before rendering
* Hydrate the config form using the full agent detail (including `system_prompt` and identity metadata)
* Add helper (`buildConfigForm`) to consistently populate form fields from agent data

## Result
* `system_prompt` is now correctly loaded and editable in the UI
* Identity fields (emoji, color, archetype, vibe, etc.) are properly populated
* Editing an agent no longer shows incomplete or stale data


jaberjaber23:
Reviewed — the `buildConfigForm` extraction is a proper refactor, and loading the `/api/agents/:id` detail **before** opening the modal (rather than setting `detailAgent` first and firing the fetch asynchronously) fixes the form-hydration race correctly. The new backend fields (`system_prompt`, `archetype`, `vibe`, `greeting_style`) are only additive on the response side.

Please rebase on latest `main` so CI runs. Good to merge once green.

PR #997, opened 2026-04-06
fix: add native Gemini embedding driver

chethanuk:
This pull request introduces support for Gemini embedding models by adding API key auto-detection for Gemini and Google, implementing a new GeminiEmbeddingDriver, and configuring default model settings. It also includes minor refactorings for MCP transport configuration and formatting cleanups in the web fetch and search modules. Feedback was provided regarding the use of unwrap_or_default() when reading response bodies, which could mask underlying network or API errors.

## Summary
- add a native Gemini embedding driver for memory embeddings
- detect when Gemini batch embedding access is unavailable and fall back to single embedContent requests
- auto-detect Gemini embedding credentials and default model selection
- update MCP transport config usage for rmcp 1.3 compatibility

## Validation
- cargo fmt --all -- --check
- cargo test --workspace embedding --quiet
- cargo clippy --workspace --all-targets -- -D warnings
- cargo build --release -p openfang-cli
- live local daemon verification with Gemini embeddings enabled
- live verification of batch-to-single fallback behavior

## **New Features**
  - Added Google Gemini as a supported embeddings provider with automatic API key detection (GEMINI_API_KEY and GOOGLE_API_KEY).
  - Updated default embedding model for Gemini/Google to a higher-capacity embedding.

* **Bug Fixes / Reliability**
  - Improved embedding request behavior with automatic fallback from batch to single requests for greater robustness.

* **Tests**
  - Simplified/clarified unit test assertions for readability.

* **Refactor**
  - Minor code and configuration cleanups for HTTP client setup and formatting.

PR #1052, opened 2026-04-14
feat(embedding): add native Google Gemini embedding driver

minhnvc:
## Summary

- Adds `GeminiEmbeddingDriver` using the `v1beta` `batchEmbedContents` API (not OpenAI-compatible — uses `x-goog-api-key` header)
- Adds `gemini-embedding-001` (3072 dims) and `text-embedding-004` (768 dims) to `infer_dimensions`
- Extracts `warn_external_api()` helper to deduplicate the locality-check warning that was copy-pasted between the Gemini and OpenAI-compat branches of `create_embedding_driver`
- Promotes batch size to module-level `GEMINI_BATCH_LIMIT` const with doc comment
- Hoists URL construction outside the chunk loop to avoid redundant allocations

## Config

```toml
[memory]
embedding_provider = "gemini"
embedding_model = "gemini-embedding-001"
embedding_api_key_env = "GEMINI_API_KEY"
```

## Test plan

- [x] `cargo build --workspace --lib` — clean compile, 0 new warnings
- [x] `test_create_embedding_driver_gemini_requires_api_key` — missing key returns `MissingApiKey`
- [ ] Live test: start daemon with `GEMINI_API_KEY` set, send agent message, verify no embedding warning in logs

### DIFF
--- a/crates/openfang-kernel/src/kernel.rs
+++ b/crates/openfang-kernel/src/kernel.rs
@@ -5908,6 +5908,7 @@ fn default_embedding_model_for_provider(provider: &str) -> &'static str {
         "together" => "togethercomputer/m2-bert-80M-8k-retrieval",
         "fireworks" => "nomic-ai/nomic-embed-text-v1.5",
         "cohere" => "embed-english-v3.0",
+        "gemini" => "gemini-embedding-001",
         // Local providers use nomic-embed-text as a good default
         "ollama" | "vllm" | "lmstudio" => "nomic-embed-text",
         // Other OpenAI-compatible APIs typically support the OpenAI model names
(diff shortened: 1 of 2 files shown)


jaberjaber23:
Thanks @minhnvc â€” the Gemini embedding driver is something we want, but after comparing with #997 (which landed slightly earlier as a parallel effort) we're going to land #997 as the canonical implementation because it adds:

- `batch â†’ single embedContent` fallback (Gemini's `batchEmbedContents` isn't available in every region/plan)
- `GEMINI_API_KEY` / `GOOGLE_API_KEY` env both accepted
- `google` alias alongside `gemini` in `create_embedding_driver`

Your implementation is cleaner and smaller, and the `GEMINI_BATCH_LIMIT = 100` constant + `x-goog-api-key` header + `warn_external_api()` helper are all nice touches. If you'd be willing to submit a follow-up PR adding those specific pieces on top of #997 (once it lands), that would be great.

Closing in favor of #997 â€” **not because your code is wrong**, just to avoid duplicate review rounds on two nearly-identical drivers. Really appreciate the contribution.

PR #1086, opened 2026-04-19
fix: flake.nix nativeBuildInputs for successful build on NixOS.

FrantaNautilus:
## Summary

The pull request fixes failing build of OpenFang on NixOS and enables installation on NixOS via flake.nix.

## Changes

Moved `perl` and `pkg-config` to nativeBuildInputs from buildInputs.

## Testing

Does not apply.

## Security

Does not apply.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -22,11 +22,13 @@
         ...
       }: {
         rust-project.src = lib.sources.cleanSource ./.;
+        rust-project.defaults.perCrate.crane.args.nativeBuildInputs = with pkgs; [
+          perl
+          pkg-config
+        ];
         rust-project.defaults.perCrate.crane.args.buildInputs = with pkgs; [
           clang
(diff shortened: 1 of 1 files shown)


jaberjaber23:
Thanks for the fix and the issue report (#1092)! Closing this in favor of #1063, which lands a superset of your fix:

- Same perl/pkg-config nativeBuildInputs move you proposed (this resolves the failure in your reproducer)
- Plus three additional NixOS issues for `openfang-desktop`: `wrapGAppsHook3`, `libayatana-appindicator` in buildInputs, and a `preFixup` LD_LIBRARY_PATH hook so the dlopen-only appindicator lib actually ends up in the runtime closure

Both PRs were correct; #1063 was just more complete. Issue #1092 is now resolved on main. Appreciate you flagging this!

PR #1104, opened 2026-04-22
fix: persist manifest changes on reload

pandego:
## Summary
- compare a broader manifest fingerprint when reloading persisted agents from disk
- include workspace and related reload-relevant fields so manifest edits are persisted back to the DB
- add regression coverage for changed and unchanged reload cases

## Validation
- cargo fmt --all --check
- cargo clippy --workspace --all-targets -- -D warnings
- cargo test --workspace

### DIFF
--- a/crates/openfang-kernel/src/kernel.rs
+++ b/crates/openfang-kernel/src/kernel.rs
@@ -464,6 +464,48 @@ fn append_daily_memory_log(workspace: &Path, response: &str) {
     }
 }
 
+fn manifests_differ_on_reload(
+    disk_manifest: &AgentManifest,
+    persisted_manifest: &AgentManifest,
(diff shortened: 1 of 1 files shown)


jaberjaber23:
Fixed in #1118. Both PRs targeted the same boot-time TOML-vs-DB diff in kernel.rs. Went with the smaller, surgical fix because of one specific concern in this approach: the JSON fingerprint in manifests_differ_on_reload treats workspace=Some(auto-assigned-path) (DB) vs workspace=None (TOML when user did not set it) as a difference and would overwrite the kernel-assigned default with None on reload. #1118 guards that case explicitly so agents that omit workspace in agent.toml keep their assigned path. Thanks for the contribution and the test coverage idea.

PR #1131, opened 2026-04-29
feat(channels): channel_id binding match rule + lenient binding parsing

### LINKED ISSUE #1127: BindingMatchRule: no `channel_id` field exists; channel-scoped routing unsupported
> **Correction (2026-04-28):** The original framing of this issue was wrong. `peer_id` is **not** overloaded with the Discord channel ID — it correctly matches the Discord **user ID**, by virtue of `bridge.rs:674-680` (`sender_user_id()`) preferring `message.metadata[\"sender_user_id\"]` (set to the author ID at `discord.rs:641`) over `sender.platform_id`. The real problem is the inverse: there is **no binding field that matches a Discord channel at all**. Issue rewritten below.

## Problem

`BindingMatchRule` (`crates/openfang-types/src/config.rs:744-758`) exposes:

```rust
pub struct BindingMatchRule {
    pub channel: Option<String>,      // platform: \"discord\", \"slack\", ...
    pub account_id: Option<String>,   // openfang account
    pub peer_id: Option<String>,      // platform user ID (for Discord: author ID)
    pub guild_id: Option<String>,     // Discord guild
    pub roles: Vec<String>,
}
```

There is **no `channel_id` field**. As a result, you cannot write a binding that says \"route messages in Discord channel X to agent Y.\" The only routing dimensions available are platform, OpenFang account, platform user, and guild.

## Why this matters

Real routing scenarios that are currently impossible to express:

- \"Messages from user A in channel #medical → researcher-A-medical; from same user A elsewhere → assistant-A.\"
- \"Anyone posting in #ops → ops-agent\" (channel-keyed routing independent of user).
- \"DMs from user X → personal-assistant-X; X's messages in shared guild channels → general-assistant.\" (Distinguishing DM from guild channel by ID requires `channel_id`.)

The information is already in the message — `sender.platform_id` carries the channel/conversation ID for Discord and Slack (per `discord.rs:647` and the comment at `bridge.rs:671-672`). It just isn't surfaced as a binding match dimension.

## Proposed fix

1. Add `channel_id: Option<String>` to `BindingMatchRule`.
2. Add `channel_id: Option<String>` to `BindingContext`.
3. Discord/Slack bridges populate `ctx.channel_id` from `sender.platform_id` (which already holds it).
4. Specificity weight for `channel_id` = 8 (peer-equivalent), so `channel_id + peer_id` (16) cleanly beats either alone (8).
5. Add `#[serde(deny_unknown_fields)]` to `BindingMatchRule` — independent of the new field, this prevents the next typo (`channnel_id`, `chan_id`, etc.) from silently producing a wide-open binding. **Highest-leverage line in the patch.**

## Spec

Detailed spec lives in `~/.openfang/workspaces/assistant/projects/openfang_channel_id_binding_spec.md` (Annabelle's workspace). Open questions to resolve before PR:

- Should the validator *warn* or *error* when `channel_id` is set without `channel`?
- Telegram / other adapters: punt to a metadata key initially, or require all adapters to populate `ctx.channel_id` from day one?

## Repro of the gap

```toml
[[bindings]]
agent = \"researcher-autumn-medical\"
match_rule = { channel = \"discord\", channel_id = \"1477803840265781391\" }
```

Today: `channel_id` is an unknown field; serde silently drops it; the binding matches every Discord message. After the fix: this works as written, and any future unknown field fails loudly at config load.

## Related

Same gap likely affects Slack channel routing and Telegram chat routing. Audit all platform adapters when implementing.

benhoverter:
## Summary

- **Feature:** Adds `channel_id` to `BindingMatchRule` so operators can route specific Discord channels / Slack conversations / Telegram chats / Matrix rooms to dedicated agents from `config.toml`. Closes #1127.
- **Resilience fix:** A typo in any single binding no longer drops the entire bindings table. Each entry is parsed independently; malformed entries log an ERROR with index + agent name, surviving entries continue to load.
- Two commits, reviewable independently:
  - `ee503b3` — `feat(channels): channel_id binding match rule`
  - `ebd05fb` — `fix(kernel): lenient binding parsing — partial-success table parse`

## Why both in one PR

The resilience fix was discovered while live-testing the `channel_id` feature (test 2: a `cccchannel_id` typo silently nuked every binding in the table). Shipping the feature without the fix would mean the very feature this PR adds is the most likely thing to silently break someone's whole routing table on first use. They belong together; the commits stay distinct so each can be reviewed on its own merits.

## Feature: `channel_id` match rule

Previously bindings keyed only on `peer_id` (user) and `guild_id` (server), forcing operators to either run multiple bots or live with one agent per server. `channel_id` closes the gap with no new adapter surface:

```toml
[[bindings]]
agent = "researcher-medical"
match_rule = { channel = "discord", channel_id = "1498782932926595293" }

[[bindings]]
agent = "researcher-business"
match_rule = { channel = "discord", channel_id = "1498782966334095553" }
```

**Data model** (`openfang-types/src/config.rs`)
- `channel_id: Option<String>` added to `BindingMatchRule`. Specificity weight = 8 (equal to `peer_id`); both together stack to 16.
- `#[serde(deny_unknown_fields)]` on `AgentBinding` and `BindingMatchRule` so typos fail loudly. Top-level `KernelConfig` stays permissive — see source comment.
- New const `CHANNELS_WITH_PLATFORM_ID_AS_CHANNEL` (19 adapters: discord, slack, telegram, matrix, mattermost, teams, webex, rocketchat, nextcloud, pumble, revolt, guilded, feishu, lark, keybase, google_chat, line, twist, flock, twitch). Single source of truth shared between routing and config validation. Hybrid adapters (IRC, Zulip) are intentionally excluded.
- Startup validation warns when `channel_id` is set for a non-supporting adapter or without a `channel`.

**Routing** (`openfang-channels/src/{types,router,bridge}.rs`)
- New `ChannelMessage::channel_id()` accessor: reads `platform_id` for allowlisted adapters, falls back to `metadata["channel_id"]`, else `None`. Case-folds `Custom(...)` variants.
- `BindingContext` gains `channel_id: Option<String>`. Bridge's `binding_context_for(message)` populates it; both dispatch paths (text + blocks) now route through `resolve_with_context`.
- Legacy `AgentRouter::resolve()` keeps `channel_id = None` — it has no inbound message. `channel_id`-scoped bindings match only through the bridge path, documented in source.

## Resilience: lenient binding parsing

`crates/openfang-kernel/src/config.rs` — new `lenient_extract_bindings` runs after include-merge / `[api]` migration, before `try_into::<KernelConfig>()`. Walks `bindings` entry-by-entry, attempts to parse each one, logs ERROR with `binding_index`, `agent`, and the underlying serde error for failures, then replaces the array with survivors. Per-entry `deny_unknown_fields` is preserved — typos still fail loudly, just no longer catastrophically. A single trailing WARN summarizes `dropped=N survivors=M`.

A broader follow-up — extending the same partial-success philosophy to the remaining `warn!` fallback in `load_config` (silent default-on-deser-failure for non-binding fields) — is left as a TODO marker in source.

## Tests

**Live-daemon validation (channel_id feature):**

| # | Test | Result |
|---|------|--------|
| 1 | Happy path: `channel + channel_id` routes to medical researcher | PASS |
| 2 | Typo guard: `cccchannel_id` rejected by `deny_unknown_fields` | PASS (surfaced the table-wide-drop bug) |
| 3 | §5.5 warning: `channel_id` without `channel` loads but warns | PASS |

**Live-daemon validation (lenient parsing):**
Typo'd `hannel` field on binding #2 — ERROR logged with index, agent name, and expected fields; remaining 5 bindings loaded and routed correctly. Single trailing WARN summarized `dropped=1 survivors=5`.

**Unit tests added:**
- *Channel_id config:* typo rejection, happy-path parse, specificity scoring (8 alone, 16 with peer_id).
- *Channel_id routing:* channel_id-only match; mismatch returns None; ctx.channel_id = None never matches a channel_id rule; channel_id + peer_id combined; three-binding scenario where channel-specific bindings beat a peer-only fallback.
- *`ChannelMessage::channel_id()`:* case-insensitivity for `Custom` variants including the Lark/Feishu Intl spelling.
- *`binding_context_for`:* Discord, Telegram, Matrix, custom-supported (Twitch), user-ID-only adapter (Reddit returns None), metadata fallback, guild_id metadata extraction, Email returns None.
- *Lenient parsing (7 tests):* main reproducer, happy-path no-op, all-malformed yields empty but rest of config loads, no-bindings-section no-op, missing-agent-field dropped, **survivor-order preservation** (locks in first-match-wins routing semantics under drop), **top-level field typo dropped** (locks in `deny_unknown_fields` at the binding root, not just inside `match_rule`).

`cargo test -p openfang-kernel config::` → 18/18 green.
Workspace `cargo check` → clean.

## Docs

`docs/channel-adapters.md` routing section rewritten: bindings are now step 1 in the resolution order. New "Bindings" subsection covers rule shape, the `peer_id` vs `channel_id` distinction, the full specificity table, the adapter allowlist with the metadata escape hatch, and the strict-field parsing rule.

## Test plan for reviewers

- [ ] `cargo test -p openfang-kernel` — config tests green
- [ ] `cargo test -p openfang-channels` — router / bridge / types tests green
- [ ] `cargo test -p openfang-types` — config tests green
- [ ] Local config with two `channel_id` bindings on the same Discord server → each channel routes to its own agent
- [ ] Introduce a single typo (e.g. `channnel_id`) in one binding → that binding logs ERROR, others continue routing
- [ ] Confirm `peer_id`-only legacy bindings still resolve unchanged

### DIFF
--- a/crates/openfang-api/src/routes.rs
+++ b/crates/openfang-api/src/routes.rs
@@ -7698,6 +7698,7 @@ pub async fn test_provider(
             Some(base_url)
         },
         skip_permissions: true,
+        subprocess_timeout_secs: None,
     };
 
     match openfang_runtime::drivers::create_driver(&driver_config) {
--- a/crates/openfang-api/src/routes.rs
+++ b/crates/openfang-api/src/routes.rs
@@ -7531,6 +7531,7 @@ pub async fn set_provider_key(
                     model: model_id,
                     api_key_env: env_var.clone(),
                     base_url: None,
+                    subprocess_timeout_secs: None,
                 };
                 let mut guard = state
                     .kernel
--- a/crates/openfang-api/tests/api_integration_test.rs
+++ b/crates/openfang-api/tests/api_integration_test.rs
@@ -61,6 +61,7 @@ async fn start_test_server_with_provider(
             model: model.to_string(),
             api_key_env: api_key_env.to_string(),
             base_url: None,
+            subprocess_timeout_secs: None,
         },
         ..KernelConfig::default()
     };
@@ -820,6 +821,7 @@ async fn start_test_server_with_auth(api_key: &str) -> TestServer {
             model: "test-model".to_string(),
             api_key_env: "OLLAMA_API_KEY".to_string(),
             base_url: None,
+            subprocess_timeout_secs: None,
         },
         ..KernelConfig::default()
     };
--- a/crates/openfang-api/tests/daemon_lifecycle_test.rs
+++ b/crates/openfang-api/tests/daemon_lifecycle_test.rs
@@ -98,6 +98,7 @@ async fn test_full_daemon_lifecycle() {
             model: "test".to_string(),
             api_key_env: "OLLAMA_API_KEY".to_string(),
             base_url: None,
+            subprocess_timeout_secs: None,
         },
         ..KernelConfig::default()
     };
@@ -225,6 +226,7 @@ async fn test_server_immediate_responsiveness() {
             model: "test".to_string(),
             api_key_env: "OLLAMA_API_KEY".to_string(),
             base_url: None,
+            subprocess_timeout_secs: None,
         },
         ..KernelConfig::default()
     };
--- a/crates/openfang-api/tests/load_test.rs
+++ b/crates/openfang-api/tests/load_test.rs
@@ -42,6 +42,7 @@ async fn start_test_server() -> TestServer {
             model: "test-model".to_string(),
             api_key_env: "OLLAMA_API_KEY".to_string(),
             base_url: None,
+            subprocess_timeout_secs: None,
         },
         ..KernelConfig::default()
     };
--- a/crates/openfang-api/tests/skill_config_api_test.rs
+++ b/crates/openfang-api/tests/skill_config_api_test.rs
@@ -76,6 +76,7 @@ async fn start_test_server() -> TestServer {
             model: "test-model".to_string(),
             api_key_env: "OLLAMA_API_KEY".to_string(),
             base_url: None,
+            subprocess_timeout_secs: None,
         },
         ..KernelConfig::default()
     };
--- a/crates/openfang-channels/src/bridge.rs
+++ b/crates/openfang-channels/src/bridge.rs
@@ -4,7 +4,7 @@
 //! `BridgeManager` which owns running adapters and dispatches messages.
 
 use crate::formatter;
-use crate::router::AgentRouter;
+use crate::router::{AgentRouter, BindingContext};
 use crate::types::{
     default_phase_emoji, AgentPhase, ChannelAdapter, ChannelContent, ChannelMessage, ChannelUser,
     LifecycleReaction,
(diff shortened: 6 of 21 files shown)


benhoverter:
Closing — branch contains 6 unrelated commits from local merges (router user-id keying + subprocess timeout) that aren't in main. Will reopen with a clean two-commit branch.

PR #1151, opened 2026-05-03
runtime/claude_code: materialize image blocks to tmpfile + extract image_cache module

benhoverter:
# PR #1151 — Updated Body (refresh of existing OPEN PR)

**Title (unchanged):** `runtime/claude_code: materialize image blocks to tmpfile + extract image_cache module`

**Base:** `main` (upstream) · **Head:** `feat/runtime-image-cache`

---

## Summary

Makes inbound image `ContentBlock`s viewable by the Claude Code CLI driver, which cannot fetch URLs or read in-memory bytes — it can only `Read` files on disk. Images are now materialized to a content-addressed tmpfile under `$HOME/.openfang/tmp/images/`, the directory is granted to the CLI via `--add-dir`, and the cache is extracted into a sibling module for reuse by outbound channel adapters.

Ten commits, no merge commits. CI runs are pending a fresh push; the prior failures (Windows-test ACL + Format) are addressed by commits 9 and 10 in the list below.

## Validation evidence

End-to-end confirmation captured 2026-05-22 against `local-main`:

- Inbound Discord image (~120 KB JPEG) → adapter emitted `Multipart([Text, Image{url}])` → bridge downloaded and cached to `~/.openfang/tmp/images/<hash>.jpg` → `claude_code` driver rendered the placeholder with the tmpfile path → CLI `Read` returned pixels → model described image contents accurately.
- Full daemon-log signature observed: `dispatch_message Multipart` → `download_image_to_blocks start` → `download_image_to_blocks complete` → `render_content Image` → `render_content materialized image to /tmp/...`.

## Commits

1. **`3525a70` runtime/claude_code: materialize images to tmpfile so the CLI's Read tool can view them**
   Adds tmpfile materialization in the `claude_code` driver. Introduces `ContentBlock::Image::source_url: Option<String>` in `openfang-types` and threads it through every consumer (api, channels/bridge, all four model drivers for pattern-match exhaustiveness, agent_loop, compactor). Field is wired through serde and matches but is not consumed by any driver on this branch — it is the load-bearing schema change for outbound Discord (separate PR). Regression test `bridge::tests::download_image_to_blocks_populates_source_url` asserts `source_url` is populated on the HTTP fetch path.

2. **`e1d694f` fix(runtime/claude_code): atomically publish materialized image tmpfiles**
[...]

Issue #1, opened 2026-02-26
guidance requested


siimaus:
Thanks for such well structured and documented project.

However at least one thing remains unclear from documents - say one wants to have agents work on repo. Where can you say to orchestrator agent - here is repo, here is issue in github, here is local folder you should work in, help me  solve the issue.

How are workspaces/files assigned/associated with agents/workflows?

Thanks

jaberjaber23:
Great question! Here's how workspace/file association works in OpenFang:

**Each agent gets its own workspace directory** — automatically created at `~/.openfang/workspaces/{agent-name}-{id}/` when spawned. You can override this in the agent manifest:

```toml
# agent.toml
workspace = "/path/to/your/repo"
```

**Agents have file tools** (`file_read`, `file_write`, `file_list`, `file_search`) that operate within their workspace. The workspace sandbox prevents agents from accessing files outside their designated directory.

**For your use case** (agent working on a GitHub repo):

1. Create an agent with the repo as workspace:
```toml
name = "repo-worker"
workspace = "/home/user/my-repo"
[capabilities]
tools = ["file_read", "file_write", "file_list", "file_search", "shell_exec", "web_fetch"]
```

2. Send it the issue context:
```bash
openfang chat --agent repo-worker
> Here is GitHub issue #42: [description]. The repo is in your workspace. Please investigate and propose a fix.
```

The orchestrator agent can decompose complex tasks and delegate to specialized agents (coder, test-engineer, etc.), each with access to the same workspace.

We'll add better documentation on this workflow. Thanks for the feedback!

PR #1143, opened 2026-04-30
fix(channels/discord): surface image attachments to text-only providers

benhoverter:
## Summary

Fixes #1142.

Discord image attachments were silently dropped before reaching the model on text-only providers. Captioned images had attachments discarded (caption-wins path); bare-image messages were dropped entirely (early `content.is_empty()` return). Result: hallucinated acknowledgements of unseen images, or no response at all.

This PR rewires the inbound attachment path end-to-end so attachments survive parsing, dispatch coherently across vision-capable and text-only providers, and render as a clear marker on text-only providers instead of being silently elided.

## Changes

- **`openfang-channels` types:** new `ChannelContent::Multipart(Vec<ChannelContent>)` variant for caption + attachment(s) as sibling blocks. Nesting forbidden (doc + `debug_assert`).
- **Discord parser:** rewritten to handle bare-image and captioned-image shapes correctly. MIME classification with filename-extension fallback for payloads missing `content_type`. 5 MB vision-size cap matching Anthropic's image-block limit; over-cap images classify as `File`. Emits `Multipart` whenever text + attachments coexist or multiple attachments are present.
- **Bridge:** flat-maps `Multipart` in both dispatch paths — `Vec<ContentBlock>` for multimodal-capable providers, newline-joined text descriptor for text-flatten providers.
- **Telegram channel:** exhaustive-match parity for the new variant; defensive outbound flatten.
- **`claude_code` driver:** renders `Image` blocks as `[attachment: <mime> image, ~N KB — not viewable on this provider]` instead of dropping them. The model still cannot see the image, but it can acknowledge it coherently rather than confabulating.

Out of scope (follow-ups): persisting attachments to disk; vision-provider dispatch refinements beyond exposing existing image bytes; non-Discord inbound attachment classification; handling for files larger than the 5 MB vision cap.

## Testing

- [x] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [x] `cargo test --workspace` passes
- [x] Live integration tested (if applicable)

Live smoke test on a `local-main` daemon build, both shapes:
- Captioned image (`"Deployed..."` + PNG) → daemon log shows `Multipart([Text, Image{...}])`; model prompt contains caption + `[attachment: image/png image, ~3172 KB — not viewable on this provider]`; LLM acknowledgement coherent.
- Bare image (PNG, no body) → daemon log shows image dispatch; model prompt contains `[attachment: image/png image, ~2774 KB — not viewable on this provider]`; LLM acknowledgement coherent.

[...]

Issue #1097, opened 2026-04-20
Unintended behavior of OpenFang workspaces


FrantaNautilus:
### Description

When an agent is given a workspace outside of openfang directory, the openfang creates agent directories in the destination. This makes workspace effectively useless, e.g. users Documents folder cannot be agent workspace. The next implication is that multiple agents cannot be pointed to a shared workspace to collaborate and exchange files. This behavior is not documented and seems to be unintended.

### Expected Behavior

Allow user to set the workspace as actual agent workspace.

### Steps to Reproduce

1. set workspace in agent.toml
2. start openfang

### OpenFang Version

0.6.0

### Operating System

Linux (x86_64)

### Logs / Screenshots

Does not apply.

jaberjaber23:
Confirmed. The root cause is in `crates/openfang-kernel/src/kernel.rs` (line 1543-1552):

```rust
let workspace_dir = manifest
    .workspace
    .clone()
    .unwrap_or_else(|| self.config.effective_workspaces_dir().join(&name));
ensure_workspace(&workspace_dir)?;
if manifest.generate_identity_files {
    generate_identity_files(&workspace_dir, &manifest);
}
```

When you set `workspace = "/home/me/Documents"` in agent.toml, the kernel uses that path as-is, then:

1. `ensure_workspace` (line 292) creates 6 subdirs at the top level: `data/`, `output/`, `sessions/`, `skills/`, `logs/`, `memory/`, plus an `AGENT.json`.
2. `generate_identity_files` (line 315) writes 7-8 .md files at the top level: `SOUL.md`, `USER.md`, `TOOLS.md`, `MEMORY.md`, `AGENTS.md`, `BOOTSTRAP.md`, `IDENTITY.md`, and `HEARTBEAT.md` for autonomous agents.

So pointing at `~/Documents` would dump all of this directly into Documents. Pointing two agents at the same shared workspace would also collide on `SOUL.md`, `AGENT.json`, and per-agent session/memory state.

## Why this needs a refactor rather than a one-line fix

The runtime depends on workspace-relative paths in many places:
- Session JSONL mirror writes to `workspace/sessions/` (kernel.rs:2230, 2796)
- Daily memory log writes to `workspace/memory/` (kernel.rs:458, 3068)
- Identity files are read for the system prompt (kernel.rs:2039-2087)
- Image generation outputs to `workspace/output/` (image_gen.rs:112)
[...]

PR #1162, opened 2026-05-04
feat(channels/discord) Outbound file/image attachments + image_cache hardening

benhoverter:
## PR-C: Outbound file/image attachments + image_cache hardening

**Branch:** `feat/outbound-file-attach` → `upstream/main`
**Stack:** independent of PR-A (#1143) and PR-B; rebased on `upstream/main` so it can land in any order. 27 commits, +4,141 / -64 across channels, runtime, and types.

### Summary

Lets agents **send images and files outbound** through the Discord channel — and tightens the inbound image cache that feeds Claude Code's Read tool.

Two concerns, one branch (kept stacked here per maintainer preference to avoid base-branch profusion):

1. **Outbound attachments** (`channels/bridge`, `channels/discord`, new `outbound_attach.rs`)
   - New `<openfang:attach path="..." />` / `url="..." />` parser in the bridge — agents emit it inline; the bridge extracts and routes to channel-specific outbound paths.
   - Discord outbound now handles `ChannelContent::File`, `Image URL`, and a true **Multipart arm** that batches N attachments into a single Discord message via `multipart/form-data` with greedy byte-cap packing (Discord's 25 MB / 10-file limits).
   - SSRF guard with redirect revalidation, scrubbed logs, content-length / streaming download cap aborts.
   - Body-aware 429 retry-after honored on the multi-file path; partial-send failures emit a structured grep-friendly WARN.
   - Parallel fetch via `try_join_all`; `Fetcher` trait extracted so tests can stub.

2. **Inbound image_cache hardening** (`runtime/image_cache`, `runtime/claude_code`, `runtime/compactor`)
   - Extracted `image_cache` into its own module (was inline in `claude_code.rs`).
   - **Filename-retaining tmpfile names:** `<hash16>__<sanitized_name>.ext` so `tmp/images/` is greppable instead of an opaque-hash junkyard. Cache-hit dedup still works via `<hash>*.<ext>` glob — first-name-wins on rename collisions.
   - Atomic publish (rename-from-temp), mtime refresh on cache hit, `--add-dir` plumbed so Claude Code's Read tool can actually open the materialized files.
   - Compactor now preserves `http(s)` `source_url` across compaction so re-materialization works after compaction events.

### Out of scope (intentional)

- Telegram / WhatsApp outbound attachments — touched only for type-signature parity.
- Pruning pre-deploy bare-hash entries in `tmp/images/` — harmless coexistence; no migration needed.

### Test plan

- [x] `cargo check` clean
- [x] `cargo clippy -D warnings` clean (only a transitive `imap-proto v0.10.2` future-incompat note, not ours)
- [x] `cargo test -p openfang-channels` — **532 unit + 9 integration passed, 0 failed**, including:
  - `test_multipart_outbound_multifile_429_*` (body-aware retry-after, both single-shot and aggregated paths)
  - greedy-pack chunker byte-cap test
  - streaming download size-cap abort test
  - per-block resolver dispatch in mixed Multipart
- [x] Live smoke against the `local-main` daemon (Discord channel):
  - Inbound: filename retention verified (`<hash16>__arize_app_01.png`, `<hash16>__pc_thumbnail_02.png`).
  - Cache-hit dedup: same bytes re-uploaded → single entry, mtime refreshed, no bare-hash sibling.
  - First-name-wins on rename: `Arize_app_01111.png` re-upload reused the original `__arize_app_01.png` entry.
  - Outbound `<openfang:attach .../>` round-trip succeeded for both URL and absolute-path forms.

### Notes for review

- `cargo fmt --check` reports drift on files **we didn't touch** (`kernel.rs`, `drivers/{anthropic,gemini,openai,vertex}.rs`). Same upstream-wide drift currently red on `main`. Will be addressed by the separate fmt-only PR; not a blocker for this one.
- The `Fetcher` trait + SSRF guard combo is the security-sensitive surface; flagging it for closer review.


jaberjaber23 (changes requested):
Substantial work and high quality on the security-sensitive surfaces. SSRF guard (literal IP + redirect revalidation + IPv4-mapped IPv6 + CGNAT), streaming size-cap with mid-stream abort verification, body-aware 429 retry with clamp, greedy-pack chunking, URL redaction in error logs — all real engineering. Tests are dense and meaningful (sweep + mtime refresh + chunk byte cap + 429 header-only + mixed types dispatch).

Three things blocking merge:

1. **outbound_attach allow-root is too wide and lets a prompt-injected agent exfiltrate the daemon's own secrets.** Default allow-root is canonicalized `~/.openfang/` (line ~3326 in default_allow_roots). That directory contains `~/.openfang/config.toml`, which holds API keys (Anthropic/OpenAI/Discord bot tokens/etc.). A prompt-injected agent emitting `<openfang:attach path="/Users/x/.openfang/config.toml"/>` would attach the whole config to its next Discord reply.

   Tighten to the two narrow subdirs you actually want: `~/.openfang/tmp/` and `~/.openfang/workspaces/<agent>/`. The skill registry/workspace_sandbox already enforces per-agent workspace confinement on the read side; the outbound attach path needs to mirror that, not undo it.

2. **Merge conflict with #1143 and #1147** (both merged on main since you opened). The Multipart variant work in bridge.rs/discord.rs/types.rs in this PR duplicates #1143 — should drop on rebase. The kernel.rs File construction got `mime`/`size` fields from #1143, your version needs to match.

3. **Coordination with #1151** (the source_url + image_cache extraction). That PR is the load-bearing schema change you reference. Decide: either land #1151 first then rebase this on top (cleanest), or close #1151 and let this carry the whole stack. Don't duplicate-merge.

Minor: `default_allow_roots()` checks HOME only (USERPROFILE on Windows yields empty list and fail-closed, which is correct, but the directory will never exist on Windows daemons). Fine to defer.

Fix #1, rebase past #1143/#1147, sort #1151 ordering, and this is ready.

jaberjaber23:
Closing. Two blockers, neither addressed.

1. **Security**: `default_allow_roots()` was `~/.openfang/` which contains `config.toml` (all API keys), credential vault, paired devices. Outbound Discord attachment would be allowed to read those by default. Flagged on the prior review, never fixed in this branch.

2. **Conflict**: branch is DIRTY against main (commit `2323cd5`). The image_cache extraction from #1151 plus the workspace_sandbox split in #1097 already changed adjacent code.

Path forward if you want this work: split into two smaller PRs.
- PR A: outbound text-only attachment (smaller, easier review).
- PR B: image attachment with `allow_roots` defaulting to **empty** (operator must explicitly opt in), and a hard refusal for paths under `~/.openfang/` regardless of allow_roots config.

Image dedup work from this branch was largely already landed in #1151. Worth rebasing on top of that rather than carrying parallel implementations.

PR #1259, opened 2026-06-30
fix(openfang): add test for issue #1212

### LINKED ISSUE #1212: Embedding driver hard-codes 6 cloud providers; OpenAI-compatible base_url override ignored (regression from #395 / #212 path for non-Ollama)
### Description

The embedding driver in v0.6.9 refuses to use any provider other than the six
hard-coded cloud providers, and none of the documented base_url override paths
work for the OpenAI driver. This blocks any OpenAI-compatible local embedding
server (llama.cpp `--embedding`, vLLM, LM Studio, etc.) from being used for
embedding-recall.

#395 was reported fixed in v0.3.35 for Ollama-via-[provider_urls]. The fix did
not generalize to the `openai` driver, which is the natural target for
OpenAI-compatible local servers. #212 closed without resolving the embedding
side of the same shape.

Concretely, the error string thrown when the embedding driver starts up is:

    set an API key (OPENAI_API_KEY, GROQ_API_KEY, MISTRAL_API_KEY,
    TOGETHER_API_KEY, FIREWORKS_API_KEY, COHERE_API_KEY)

That allowlist is hard-coded; there is no path to point at a self-hosted
endpoint without forking and patching.

The result for self-hosted lab deployments: chat routing works fine to local
providers (Ollama, vLLM via OLLAMA_HOST / VLLM_HOST), but embedding-recall
still goes to OpenAI cloud, leaking text-content the operator wanted to keep
in-lab and burning OpenAI credits on what should be a local-only call path.

### Expected Behavior

Either of these would unblock self-hosted deployments:

(A) Honor a per-driver `base_url` for the openai driver, so:

      [providers.openai]
      base_url = "http://my-llama-cpp:9080/v1"

    causes the embedding driver to issue POSTs against
    http://my-llama-cpp:9080/v1/embeddings instead of api.openai.com.

(B) Add `local`/`openai_compatible`/`llamacpp` as recognized embedding
    providers in the [memory] block, with their own base_url field:

      [memory]
      embedding_provider = "openai_compatible"
      embedding_model    = "nomic-embed-text-v1.5"
      embedding_base_url = "http://my-llama-cpp:9080/v1"
      embedding_api_key_env = "OPENAI_API_KEY"   # ignored if local accepts no auth

(B) is preferable because it keeps OPENAI_API_KEY meaning the public API,
and makes "this embedding call goes to my LAN" explicit in config. Either
solves the use case.

### Steps to Reproduce

1. Stand up an OpenAI-compatible embedding endpoint on a separate host, e.g.:
       llama-server -m nomic-embed-text-v1.5.gguf --embedding --port 9080
   Verify it works:
       curl -s http://my-embedding-host:9080/v1/embeddings \
            -H 'Content-Type: application/json' \
            -d '{"model":"nomic-embed-text-v1.5","input":"hello"}' \
        | jq '.data[0].embedding | length'
   → returns 768 (or whatever dim the model uses).

2. Try Approach A — env var OPENAI_BASE_URL:
       OPENAI_BASE_URL=http://my-embedding-host:9080/v1 openfang start
   Result: embedding driver still issues requests to api.openai.com.
   The env var is silently ignored.

3. Try Approach B — [providers.openai] base_url in config.toml:
       [providers.openai]
       base_url = "http://my-embedding-host:9080/v1"

       [memory]
       embedding_provider = "openai"
       embedding_model = "nomic-embed-text-v1.5"
       embedding_api_key_env = "OPENAI_API_KEY"
   GET /api/providers shows the override is stored.
   Result: embedding driver STILL issues requests to api.openai.com.
   Caches its URL at init; ignores provider table for embedding path.

4. Try Approach C — runtime override via API:
       curl -X PUT http://localhost:4200/api/providers/openai/url \
            -H 'Content-Type: application/json' \
            -d '{"url":"http://my-embedding-host:9080/v1"}'
   Result: 500 — "Read-only file system" on the persisted config write
   (in containerized deploys; not the embedding fix but worth noting that
   even the writable-config path doesn't help embeddings).

5. Try Approach D — set the [memory] block to a non-openai provider:
       [memory]
       embedding_provider = "local"   # or "llamacpp" or "openai_compatible"
       embedding_model = "nomic-embed-text-v1.5"
   Result: startup fails with the hard-coded allowlist error above.

### OpenFang Version

0.6.9

### Operating System

Linux (x86_64)

### Logs / Screenshots

$ kubectl -n openfang logs openfang-0 | grep -i embed
[INFO] memory: embedding_provider = openai, model = nomic-embed-text-v1.5
[ERROR] embedding driver: set an API key (OPENAI_API_KEY, GROQ_API_KEY,
        MISTRAL_API_KEY, TOGETHER_API_KEY, FIREWORKS_API_KEY,
        COHERE_API_KEY)

$ curl -s http://localhost:4200/api/providers | jq '.openai'
{
  "url": "http://my-embedding-host:9080/v1",   # <-- override visible
  "key_present": true
}

$ # ... but the embedding request still hits api.openai.com:
$ kubectl -n openfang exec openfang-0 -- ss -tnp
ESTAB  ...  api.openai.com:443  ...

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #1212.

## Why
Fixes #1212

## Testing
- [x] New test added


jaythehardcoder:
Closing this PR — the stub doesn't address the underlying issue. I'll follow up with a proper fix when I have a working reproduction. Thank you for maintaining this project.

PR #1261, opened 2026-07-01
fix(openfang): add test for issue #1154

### LINKED ISSUE #1154: LM  STUDIO / OLLAMA is not setup to allow for any use case
### Description

There should be the ability to use ip address's not just local host. This can be on a vps, lxc conatiner or small computer that uses a main omputer for offline llms

### Expected Behavior

Allow for all forms and easy setup (open claw has this feature now as well could be used for reference). 

Init logs are not setting up themselves to run this after reboot automatically either

### Steps to Reproduce

init >>>> lm studio local only

### OpenFang Version

any

### Operating System

Linux (x86_64)

### Logs / Screenshots

_No response_

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #1154.

## Why
Fixes #1154

## Testing
- [x] New test added


jaythehardcoder:
Closing this PR — the stub doesn't address the underlying issue. I'll follow up with a proper fix when I have a working reproduction. Thank you for maintaining this project.

PR #1267, opened 2026-07-13
Add MiniMax M3 to provider catalog

octo-patch:
## Summary

- Register MiniMax-M3 in the built-in catalog and make it the default MiniMax model across setup flows.
- Keep MiniMax-M2.7 available while correcting its output limit from the current official specification.
- Preserve M3 input-length pricing and document regional OpenAI-compatible and Anthropic-compatible configuration.

## Changes

- Add M3 context, output, vision, tool, alias, and pricing metadata.
- Apply the long-context standard rate when input exceeds 512,000 tokens.
- Add a request-capture test for the Anthropic-compatible `/v1/messages` path.
- Document both regions, both protocols, thinking modes, modalities, and all current pricing tiers.

## Testing

- [x] `cargo fmt --all -- --check`
- [x] `cargo test --workspace`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [ ] Live integration tested (not run; local request capture verifies endpoint construction without external credentials)

## Security

- [x] No new unsafe code


octo-patch:
Thanks for the review. I updated the MiniMax M3 catalog and docs pricing to use the recorded flat input/output rates and removed the unsupported tiered metering path, then pushed a new commit. I ran , , , , and .

PR #1281, opened 2026-08-31
fix(runtime): extract text from PDF in web_fetch and cap chunked responses (#1271)

### LINKED ISSUE #1271: web_fetch injects raw PDF binary into agent context instead of extracting text
## Summary

`web_fetch` decodes every HTTP response body via `resp.text()`, which turns a
PDF into a `String` of raw, FlateDecode-compressed bytes and passes it straight
into the agent's context. A single medium PDF (a restaurant menu in our case)
produced ~617 K characters of binary. This overflows the model's context window
and — with a local model — makes it emit garbled tokens that persist across the
whole thread, since the poisoned history is re-sent on every subsequent turn.

The Content-Type is read (`web_fetch.rs`, ~L115) but never used except for the
HTML check, so `application/pdf` falls through to the raw-body branch.

## Reproduction

1. Point any agent at a PDF URL and ask it to read the content, e.g.
   `web_fetch https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf`
2. Observe the tool result contains `%PDF-1.x … stream … endstream` raw bytes.
3. With a larger PDF, the request exceeds the context window
   (`exceed_context_size_error`) and following messages return corrupted output.

## Second, related issue at the same spot

The size guard at `web_fetch.rs` L106 only fires when the server sends a
`Content-Length` header. For chunked responses without it, `resp.text()`
buffers the entire body into memory unbounded before `max_chars` truncation —
a memory-exhaustion vector.

## Note on the bundled `pdf-reader` skill

OpenFang ships a `pdf-reader` skill, but it is prompt-only: it can't help here,
because the raw binary has already entered the context before any skill logic
runs. The fix has to happen at the tool layer, before the body reaches the model.

## Proposed fix

Read the body as a size-capped byte stream and route by Content-Type:
`application/pdf` → text extraction (pure-Rust `pdf-extract`, no new system deps),
everything else → lossy UTF-8. Binary is never passed through as a string, and
the streaming cap also closes the chunked-response memory issue.

We already have this implemented and running against v0.6.9 (`acf2587`):
`cargo test --workspace`, `cargo clippy -D warnings`, and `cargo fmt --check`
all pass, plus 4 new tests. **Happy to open a PR if you'd welcome it.**

Adityakk9031:
## Description
Fixes #1271.

\web_fetch\ previously decoded every HTTP response directly into string text via \
esp.text()\. When fetching PDFs, this resulted in raw FlateDecode-compressed binary byte strings entering the agent context (often hundreds of thousands of characters of binary tokens), overflowing context windows and poisoning conversation history. Additionally, chunked responses lacking a \Content-Length\ header bypassed the response size guard and were read unbounded into memory.

### Changes
1. **PDF Text Extraction**:
   - Added \pdf-extract\ (v0.8) and \lopdf\ (v0.34) workspace dependencies.
   - Added \is_pdf()\ detection matching \pplication/pdf\, \pplication/x-pdf\, and \%PDF\ magic byte headers.
   - Added \extract_pdf_content()\ to extract plain text from PDF bytes in memory, providing structured fallback descriptions if the PDF contains scanned images without OCR or is encrypted/corrupted.
2. **Chunked Streaming & Bounded Response Guard**:
   - Replaced unbounded \
esp.text()\ with chunked stream reader \
ead_bounded_body()\ that enforces \max_response_bytes\ dynamically across all chunks, protecting against memory exhaustion even without \Content-Length\ headers.
3. **Binary Content Sniffing**:
   - Added \is_binary_content()\ to sniff binary MIME types (\image/*\, \ideo/*\, \udio/*\, \pplication/zip\, \pplication/octet-stream\) and sample null bytes in the first 1024 bytes.
   - Returns structured metadata notes for binary responses instead of dumping raw binary byte strings into context.
4. **Lossy UTF-8 Fallback**:
   - Handled non-UTF8 text decoding with lossy UTF-8 conversion (\String::from_utf8_lossy\).
5. **Applied to both Engine and Legacy Tool**:
   - Updated both \WebFetchEngine::fetch\ and \	ool_web_fetch_legacy\ in \	ool_runner.rs\.
6. **Unit Tests**:
   - Added unit tests for PDF detection, binary detection, valid PDF text extraction, and corrupted PDF fallback handling.

### Testing
- Ran unit tests in \crates/openfang-runtime/src/web_fetch.rs\ (21 passed, 0 failed).
- Verified shellcheck on \scripts/install.sh\.


Adityakk9031:
@jaberjaber23 have a look

